PR #1, opened 2020-12-05
Hook up the PyPI feed to PubSub.

dlorenc:
Also a small fix to the XML handling.


feeds/pypi/main.go
```diff
@@ -58,32 +77,47 @@ func fetchPackages() ([]*Package, error) {
...
 
-// PubSubMessage is the payload of a Pub/Sub event.
-type PubSubMessage struct {
-	Data []byte `json:"data"`
-}
-
 // Poll receives a message from Cloud Pub/Sub. Ideally, this will be from a
 // Cloud Scheduler trigger every `delta`.
-func Poll(ctx context.Context, m PubSubMessage) error {
+func Poll(ctx context.Context, t *pubsub.Topic) error {
```
jordan-wright, line 82: This is the one thing I'm not sure about. I anticipated that this would work via a pubsub trigger that occurs from a cloud scheduler. Basically setting up the cloud scheduler [like this](https://cloud.google.com/scheduler/docs/tut-pub-sub) to run every 5 minutes, and then following [this tutorial](https://cloud.google.com/functions/docs/calling/pubsub) to trigger the cloud function. That's where the message signature came from, and the `main` was just for local testing.

I'm _far_ from a GCP expert so please correct me, but it looks like this adjusts things to be more of a long-lived process that pulls from a topic? Do we anticipate that we can use the changes you've made here to do the time-based execution I mentioned above?

feeds/pypi/main.go
```diff
@@ -17,16 +27,25 @@ type Response struct {
...
 	Link         string      `xml:"link"`
-	Name         string      `xml:"-"`
-	Version      string      `xml:"-"`
 }
 
 type rfc1123Time struct {
 	time.Time
 }
 
+func (p *Package) Name() string {
```
jordan-wright, line 39: I'm good with this! One of the things that took me a bit was that I originally tried to be clever and do a custom unmarshaler to split the link, but that turned into a self-referential infinite loop, which would have required a type alias, etc. It was all very gross so I figured "eh, let's just parse it then split it" and this follows that same idea. So 👍 from me!

feeds/pypi/main.go
```diff
@@ -58,32 +77,47 @@ func fetchPackages() ([]*Package, error) {
...
+		b, err := json.Marshal(msg)
+		if err != nil {
+			return err
+		}
+		if err := t.Send(ctx, &pubsub.Message{
+			Body: b,
+		}); err != nil {
+			return err
+		}
+		log.Println("Sent")
```
jordan-wright, line 107: Small nit: should we consolidate the first log message and this one so that we have a single message we can review that says "sent <package> <version>"? I feel like that'll make debugging easier. Totally fine if we want to keep the first message around so that we can see both what packages we've seen, and which we've sent upstream.

feeds/pypi/main.go
```diff
@@ -58,32 +77,47 @@ func fetchPackages() ([]*Package, error) {
...
-func Poll(ctx context.Context, m PubSubMessage) error {
+func Poll(ctx context.Context, t *pubsub.Topic) error {
 	packages, err := fetchPackages()
 	if err != nil {
 		return err
 	}
 	cutoff := time.Now().UTC().Add(-delta)
 	for _, pkg := range packages {
+		log.Println(pkg.Name(), pkg.Version())
 		if pkg.ModifiedDate.Before(cutoff) {
```
jordan-wright, line 90:
```suggestion
		if pkg.ModifiedDate.After(cutoff) {
```

This was another bug I spotted during testing. Figured we could just make it part of this PR if you'd like.

feeds/pypi/main.go
dlorenc: > I'm _far_ from a GCP expert so please correct me, but it looks like this adjusts things to be more of a long-lived process that pulls from a topic?

Ah, no. This topic is the thing we *write to* after processing the index.

We still need to hook up the thing that "triggers" this to run after some kind of cron.

feeds/pypi/main.go
dlorenc: Whoops, meant to drop these. I added them when debugging and got confused.

feeds/pypi/main.go
dlorenc: Ah interesting. Yeah I was getting empty values and this was the fastest way I could think of. The alias/infinite recursion problem makes sense now though. Thanks for saving me the time!

feeds/pypi/main.go
dlorenc: Fixed to keep the signature!

feeds/pypi/main.go
```diff
@@ -66,24 +85,43 @@ type PubSubMessage struct {
 // Poll receives a message from Cloud Pub/Sub. Ideally, this will be from a
 // Cloud Scheduler trigger every `delta`.
 func Poll(ctx context.Context, m PubSubMessage) error {
```
jordan-wright, line 87: @dlorenc one more note on this one. It looks like to move to Cloud Run, we need to expose a standard http handler. In my fork, I was able to make this just the standard `Poll(w http.ResponseWriter, r *http.Request)`, then triggering it directly from cloud scheduler. If we can make that small adjustment, then we'll be in good shape to merge.

jordan-wright:
Merging this since I've converted it to support Cloud Run and it should work 😄

PR #2, opened 2020-12-06
Initial commit of Terraform deployment workflow

jordan-wright:
This PR aims to add support for handling automatic deployments via Terraform and GitHub Actions. I plan to do some testing on a local fork, but I wanted to put the PR up earlier than later to start collecting feedback.

The Terraform deployment creates a Cloud Scheduler that runs every 5 minutes, triggering a Pub/Sub topic which, in turn, triggers a function that runs our PyPI polling.

The function is created and expects the code to be provided in a cloud storage bucket. I've included a GitHub Actions workflow that would run on every push to `master`, building the function archive then running terraform to apply any latest changes, keeping the state in a storage bucket.

As always, open to thoughts/suggestions!

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,7 +24,7 @@ jobs:
 
     - name: Setup Terraform
       uses: hashicorp/setup-terraform@v1
-    
+
     - name: Setup Credentials
       run: echo ${{ secrets.TF_GOOGLE_CREDENTIALS }} | base64 -d > /credentials.json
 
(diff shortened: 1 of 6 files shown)


jordan-wright:
The next steps include local testing on a fork first, then making a deployment service account in the project and adding the credentials as a secret.

dlorenc:
Nice! This looks good to me. CI seems to be having trouble though.

jordan-wright:
I'm going to work to adjust this per our conversation about moving away from Cloud Functions and to Cloud Run. Shouldn't be massive changes, but it'll take some experimentation.

And fwiw CI is failing because I haven't set up the service account in GCP yet. Once we do that we'll be in good shape. Just waiting until we have a good deployment story 😄

jordan-wright:
Looks reasonable to me. Gonna merge 

🤞 things apply successfully.

PR #6, opened 2021-01-30
Hookup a basic scheduler to consume tasks from the cloud run pubsub q…

dlorenc:
…ueue.

This runs in kubernetes now because it will eventually need to create k8s pods.
Creating from inside the cluster is easier than allowing access from outside.


dlorenc:
Right now this just logs the payload. Up next is time to deserialize and create pods!

PR #10, opened 2021-02-01
Updated go.sum

naveensrinivasan:
go.sum was outdated.


naveensrinivasan:
@dlorenc I guess the secret is not accessible because I don't have access to the repo and secrets aren't available in forks.

dlorenc:
Yup, thanks!

PR #18, opened 2021-02-02
Setup infrastructure for the npm feed

jordan-wright:
This PR sets up the Terraform and Cloud Run scaffolding for the npm feed introduced in #16.

PR #20, opened 2021-02-02
Fix npm path when building Cloud Run images

jordan-wright:
Looks like PR #18 was merged with a small error when building cloud images. This should be fixed now. 😄 

(Note: I want to make sure that checks pass before merging to catch any other issues)


jordan-wright:
This looks better - we should be good to merge! 🎊

PR #28, opened 2021-02-03
Switch the service name to ruby.

dlorenc:


dlorenc:
Gonna merge this one, the one from before actually took down NPM and gave us two rubys...

jordan-wright:
Oh no! That's my bad. Good catch. Fortunately the refactor to bundle these will also make this not a thing in the future.

PR #39, opened 2021-02-05
Refactor scheduler job into more reusable module

steiza:
I'm a bit new to Terraform, so hopefully this is helping rather than causing more work!

At @jordan-wright's suggestion, I made a Terraform module to reduce the amount of broiler-plate needed for adding a new packaging ecosystem.

I tested this by running `terraform plan` locally (against an empty GCP account I made for testing), and it looked sane. I would strongly encourage someone to double-check this against the real infrastructure before running it.


jordan-wright:
A local `terraform plan` looks reasonable to me! And reviewing the changes, it looks like it does exactly what's suggested - moving the scheduler logic into a self-contained module.

I'll merge this in and keep an eye to make sure that the deploy works as expected.

Thanks @steiza!

PR #40, opened 2021-02-06
Feature - Pulumi for creating infrastructure 

naveensrinivasan:
`brew install pulumi` https://www.pulumi.com/docs/get-started/

Use pulumi to create `k8s` cluster, bucket, and registry.  

The goals are to make the cluster creation optional and then use pulumi to deploy analysis code and also Falco rules with helm.

The README in this PR has the documentation for setting up pulumi.


dlorenc:
Nice! I'd love to give this a try. Could you add a few more instructions either here or in a README.md in the pulumi directory on how to actually run this to setup a cluster and stuff?

naveensrinivasan:
> Nice! I'd love to give this a try. Could you add a few more instructions either here or in a README.md in the pulumi directory on how to actually run this to setup a cluster and stuff?

Absolutely! Thanks

analysis/config/deployment.yaml
```diff
@@ -19,7 +19,7 @@ spec:
         image: ko://github.com/ossf/package-feeds/analysis
         env:
           - name: OSSF_MALWARE_ANALYSIS_RESULTS
-            value: gs://ossf-malware-analysis-results
+            value: gs://naveen-analysis
```
naveensrinivasan, line 22: This is me mucking around to make sure it is able to deploy it. This will not be committed.

naveensrinivasan:
@dlorenc  Included instructions in README. Let me know if you are able to proceed with it.

naveensrinivasan:
This is how pulumi bot would comment on new PR https://github.com/naveensrinivasan/package-feeds/pull/2#issuecomment-775231377 

For this GCP token and Pulumi token has to be set as GitHub secrets, which is mentioned in README

naveensrinivasan:
@dlorenc ping

dlorenc:
Taking a look!

dlorenc:
Nice! So just to make sure I understand, a few questions:

- Right now, this doesn't actually deploy any containers in the cluster?
- Does pulumi support helm for the helm charts we use?
- I've been building and deploying with `ko` instead of Dockerfiles - would that still work?

The nice things:

- I like the idea of using the same tool to manage creating the k8s cluster and deploying into it. It seems like Terraform is a better fit for creating the infra rather than deploying into it, so this might be a nice tool for doing both.
- The workload identity setup is awesome, that's always pretty manual whenever I do it with other tools

The not-so-nice things:

- I don't love having to sign up for a Pulumi account to use this - is there a way to run it without an account?
- There's a lot of boilerplate here: the makefile, the linting, etc. Do we need all of that for the infra scripts (even though they're Golang...)?
- It's hard to see how it all fits together without seeing how the containers get built and deployed too. Or is the idea that pulumi would only be for spinning up infra and we'd use something else to deploy into it?

naveensrinivasan:
> Nice! So just to make sure I understand, a few questions:
> 
> * Right now, this doesn't actually deploy any containers in the cluster? 

Yes, right now it doesn't. It can very much do that.

> * Does pulumi support helm for the helm charts we use?

Yes, pulumi supports helm.

https://www.pulumi.com/docs/guides/adopting/from_kubernetes/#provisioning-a-helm-chart

> * I've been building and deploying with `ko` instead of Dockerfiles - would that still work?

 I would have to investigate to check if it would or not. How critical is this?

> 
> The nice things:
> 
> * I like the idea of using the same tool to manage creating the k8s cluster and deploying into it. It seems like Terraform is a better fit for creating the infra rather than deploying into it, so this might be a nice tool for doing both.
> * The workload identity setup is awesome, that's always pretty manual whenever I do it with other tools
> 
> The not-so-nice things:
> 
> * I don't love having to sign up for a Pulumi account to use this - is there a way to run it without an account?

I agree I wish the state was managed without signup, but pulumi manages the state in the cloud and that is the reason for the signup.


> * There's a lot of boilerplate here: the makefile, the linting, etc. Do we need all of that for the infra scripts (even though they're Golang...)?

That is to make sure the code is clean.  The next stage is to writing unit-tests. But I am OK if you think it is not necessary.

> * It's hard to see how it all fits together without seeing how the containers get built and deployed too. Or is the idea that pulumi would only be for spinning up infra and we'd use something else to deploy into it?

No, pulumi would deploy Infrastructure and the containers 

Here are a few examples

1. https://www.pulumi.com/docs/guides/crosswalk/kubernetes/app-services/
2. https://www.pulumi.com/docs/guides/crosswalk/kubernetes/control-plane/
3. https://github.com/pulumi/examples/tree/master/kubernetes-go-guestbook

PR #45, opened 2021-02-07
Add back the pubsub URL prefix.

dlorenc:
Scheduler jobs are failing with: "no scheme in URL "projects/ossf-malware-analysis/topics/feed-topic""


jordan-wright:
Good call! I missed that in my review, even though that exact thing has bitten me before. 😳

PR #47, opened 2021-02-07
Make the images we run installation in look a little more realistic.

dlorenc:
The jobs now run as a non-root user in a real home directory.

### DIFF
--- a/analysis/customrules.yaml
+++ b/analysis/customrules.yaml
@@ -33,7 +33,7 @@ customRules:
       desc: Detect file access outside the workingdir
       condition: >
         (open_read and 
-         not fd.name startswith /app/ and
+         not (fd.name startswith /app/ or fd.name startswith /home/fakeuser/app/) and
          not fd.name startswith /tmp/ and
          not fd.name startswith /usr/local/lib/ and
          not fd.name startswith /usr/lib/ and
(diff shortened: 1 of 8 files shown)


naveensrinivasan (approved):
Nice, Would distroless images like these would be better of ?https://github.com/GoogleContainerTools/distroless/blob/master/examples/python3/Dockerfile
https://github.com/GoogleContainerTools/distroless/blob/master/examples/nodejs/Dockerfile

jordan-wright:
Out of curiosity, what do we define as "realistic"? On one hand, we could say that we want these to look like a developer's laptop, but that might introduce things that are easier to fingerprint.

On the other hand, I've historically leaned towards "realistic" meaning what would likely be seen running in CI jobs at large organizations. To that end, I'd encourage keeping the Dockerfiles bland and boring, preferring the default images as-is wherever possible. That way there's no fingerprinting that can be done without also severely limiting the reach of the malware.

dlorenc:
That's fair! I was actually trying to simulate "desktop/laptop" scenarios, where you're running as non root and you're worried about personal credentials.

The CI one is another threat model where the root is way more realistic like you said.

dlorenc:
Opened https://github.com/ossf/package-feeds/issues/49 to track thinking through other ways to run these.

PR #48, opened 2021-02-07
Refactor scheduled feeds into a single Cloud Run service

jordan-wright:
This is the first pass at refactoring our existing feeds to run in a single scheduled service.

There's a number of moving parts here, and some of the difficulty is just due to the repository layout. When we split off analysis from the feeds, I anticipated this will be cleaner.

For now, here are the big changes:

* I've moved the executable bits into a single `cmd/scheduled-feed/` package. This is the entrypoint for the Cloud Run service.
* There is a new `feeds/scheduler` package that keeps a registry of feeds and exposes a single function that iterates over them, pulling the latest packages
* I've updated the Cloud Build invocation to use a config file so that I could both send the full source context up to GCP, while keeping the Dockerfile in the `cmd/scheduled-feed` folder. That way the Dockerfile isn't sitting in the root of the repo, those this will likely change when we split things out.
* I've renamed the previous `*/main.go` files to their package equivalents. That's why it looks like every feed is getting recreated.

I'd ask that we hold off on merging this for now. My plan is to apply the Terraform manually that way I can tweak things if things break without the need for more PR cycles. I'll keep Slack updated with that progress.


feeds/scheduler/scheduler.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/ossf/package-feeds/feeds/npm"
+	"github.com/ossf/package-feeds/feeds/pypi"
+	"github.com/ossf/package-feeds/feeds/rubygems"
+	log "github.com/sirupsen/logrus"
+)
+
+// ScheduledFeeds is a registry of feeds that should be run on a schedule
+var ScheduledFeeds = make(map[string]feeds.ScheduledFeed)
+
+func init() {
```
naveensrinivasan, line 17: How about avoiding the global state and avoiding the init functions? This will make it clean and don't have to worry about `mutex` in the future.  Instead of the global state use something like `NewObject` which returns the map.

Something like this 

```
func NewScheduler() map[string]feeds.ScheduledFeed {
	ScheduledFeeds := make(map[string]feeds.ScheduledFeed)
	ScheduledFeeds[pypi.FeedName] = pypi.Feed{}
	ScheduledFeeds[npm.FeedName] = npm.Feed{}
	ScheduledFeeds[rubygems.FeedName] = rubygems.Feed{}
	ScheduledFeeds[crates.FeedName] = crates.Feed{}
	return ScheduledFeeds
}

// PollScheduledFeeds fetches the latest packages from each registered feed
func PollScheduledFeeds(cutoff time.Time, scheduledFeeds map[string]feeds.ScheduledFeed) ([]*feeds.Package, error) {
	packages := []*feeds.Package{}
	for name, feed := range scheduledFeeds {
		logger := log.WithField("feed", name)
		pkgs, err := feed.Latest(cutoff)
		if err != nil {
			logger.WithError(err).Error("error fetching packages")
			return nil, err
		}
		processed := 0
		for _, pkg := range pkgs {
			processed++
			packages = append(packages, pkg)
		}
		logger.WithField("num_processed", processed).Print("processed packages")
	}
	return packages, nil
}
```

https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html
https://github.com/leighmcculloch/gochecknoinits

feeds/scheduler/scheduler.go
```diff
@@ -0,0 +1,42 @@
...
+	ScheduledFeeds[rubygems.FeedName] = rubygems.Feed{}
+	ScheduledFeeds[crates.FeedName] = crates.Feed{}
+}
+
+// PollScheduledFeeds fetches the latest packages from each registered feed
+func PollScheduledFeeds(cutoff time.Time) ([]*feeds.Package, error) {
+	packages := []*feeds.Package{}
+	for name, feed := range ScheduledFeeds {
+		logger := log.WithField("feed", name)
+		pkgs, err := feed.Latest(cutoff)
```
naveensrinivasan, line 29: What are your thoughts on running each of these in a goroutine `feed.Latest`?

feeds/npm/.golangci.yml
```diff
@@ -61,7 +61,7 @@ linters:
     - unused
     - varcheck
     - whitespace
-    - wrapcheck
+    # - wrapcheck
```
naveensrinivasan, line 64: Any reason for these to commented out?

feeds/npm/.golangci.yml
```diff
@@ -29,7 +29,7 @@ linters:
     - goheader
     - goimports
     - golint
-    - gomnd
+    # - gomnd
```
naveensrinivasan, line 32: Any reason for this to be commented out?

cmd/scheduled-feed/main.go
```diff
@@ -0,0 +1,66 @@
...
+	topicURL := os.Getenv("OSSMALWARE_TOPIC_URL")
+	topic, err := pubsub.OpenTopic(context.TODO(), topicURL)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	cutoff := time.Now().UTC().Add(-delta)
+	pkgs, err := scheduler.PollScheduledFeeds(cutoff)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
naveensrinivasan, line 31: Logging errors here similar to other error conditions would be helpful and consistent.

cmd/scheduled-feed/main.go
```diff
@@ -0,0 +1,66 @@
...
+	_ "gocloud.dev/pubsub/gcppubsub"
+)
+
+var delta = 5 * time.Minute
+
+func Poll(w http.ResponseWriter, r *http.Request) {
+	topicURL := os.Getenv("OSSMALWARE_TOPIC_URL")
+	topic, err := pubsub.OpenTopic(context.TODO(), topicURL)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
naveensrinivasan, line 25: Logging errors here similar to other error conditions would be helpful and consistent.

cmd/scheduled-feed/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/ossf/package-feeds/feeds/scheduler"
+
+	log "github.com/sirupsen/logrus"
+
+	"gocloud.dev/pubsub"
+	_ "gocloud.dev/pubsub/gcppubsub"
+)
+
+var delta = 5 * time.Minute
```
naveensrinivasan, line 19: This could be a `const`

cmd/scheduled-feed/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+func Poll(w http.ResponseWriter, r *http.Request) {
+	topicURL := os.Getenv("OSSMALWARE_TOPIC_URL")
```
naveensrinivasan, line 22: Check for `OSSMALWARE_TOPIC_URL` being empty should be handled in `main` with a panic,

feeds/feed.go
```diff
@@ -0,0 +1,14 @@
+package feeds
+
+import "time"
+
+type ScheduledFeed interface {
```
naveensrinivasan, line 5: 👍

naveensrinivasan (commented):
Like the interface which is much cleaner.

feeds/scheduler/scheduler.go
naveensrinivasan: Something like this.
```
func PollScheduledFeeds(cutoff time.Time) ([]*feeds.Package, error) {
	var err error
	type success struct {
		name   string
		result []*feeds.Package
	}

	type failure struct {
		name string
		err  error
	}
	packages := []*feeds.Package{}
	results := make(chan success, len(ScheduledFeeds))
	errorsResults := make(chan failure, len(ScheduledFeeds))

	for name, feed := range ScheduledFeeds {
		go func(name string, feed feeds.ScheduledFeed) {
			pkgs, err := feed.Latest(cutoff)
			if err != nil {
				errorsResults <- failure{name, err}
			}
			results <- success{name, pkgs}
		}(name, feed)
	}

	for i := 1; i <= len(ScheduledFeeds); i++ {
		select {
		case pkgs := <-results:
			logger := log.WithField("feed", pkgs.name)
			packages = append(packages, pkgs.result...)
			logger.WithField("num_processed", len(pkgs.result)).Print("processed packages")
		case errResult := <-errorsResults:
			logger := log.WithField("feed", errResult.name)
			logger.WithError(errResult.err).Error("error fetching packages")
			err = multierror.Append(err, errResult.err)
		}
	}
	return packages, err
}

```

jordan-wright: I'm definitely in favor of doing this! But I think I'll leave it for a separate PR/optimization to try and keep this already complex refactor to be as simple as possible.

feeds/npm/.golangci.yml
jordan-wright: Because they were catching errors that existed prior to this refactor. My goal with this PR is just to reorganize things. I'll fix linting issues with the existing code in a future PR 😄

jordan-wright:
This is in good shape - I've applied the terraform and confirmed things appear to be running.

@naveensrinivasan if you want to give this one last lookover, feel free! There are things that I'm intentionally deferring until a future PR to try and keep this as simple as possible.

Otherwise, I'll merge this in Tuesday or so. 🙂

naveensrinivasan:
@jordan-wright Thanks. Could you open tracking issues for those items?

Issue #53, opened 2021-02-21
Leverage Goroutines for Feeds


jordan-wright:
Since each feed is atomic, we can wrap each one in a goroutine and consolidate the results, making runs even quicker.

Ref: 
```
func PollScheduledFeeds(cutoff time.Time) ([]*feeds.Package, error) {
	var err error
	type success struct {
		name   string
		result []*feeds.Package
	}

	type failure struct {
		name string
		err  error
	}
	packages := []*feeds.Package{}
	results := make(chan success, len(ScheduledFeeds))
	errorsResults := make(chan failure, len(ScheduledFeeds))

	for name, feed := range ScheduledFeeds {
		go func(name string, feed feeds.ScheduledFeed) {
			pkgs, err := feed.Latest(cutoff)
			if err != nil {
				errorsResults <- failure{name, err}
			}
			results <- success{name, pkgs}
		}(name, feed)
	}

	for i := 1; i <= len(ScheduledFeeds); i++ {
		select {
		case pkgs := <-results:
			logger := log.WithField("feed", pkgs.name)
			packages = append(packages, pkgs.result...)
			logger.WithField("num_processed", len(pkgs.result)).Print("processed packages")
		case errResult := <-errorsResults:
			logger := log.WithField("feed", errResult.name)
			logger.WithError(errResult.err).Error("error fetching packages")
			err = multierror.Append(err, errResult.err)
		}
	}
	return packages, err
}

```

_Originally posted by @naveensrinivasan in https://github.com/ossf/package-feeds/pull/48#discussion_r573364318_

tom--pollard:
This can be closed now right? :)

naveensrinivasan:
Yes, Thanks.

PR #54, opened 2021-02-21
Leverage goroutines for processing feeds

jordan-wright:
This PR implements the previous suggestion in #53 to use goroutines when fetching new packages.

Worth noting I've decided to just log errors for now instead of failing closed. That way if one feed is broken, we still send results for the rest of them.

### DIFF
--- a/cmd/scheduled-feed/main.go
+++ b/cmd/scheduled-feed/main.go
@@ -24,10 +24,12 @@ type FeedHandler struct {
 
 func (handler *FeedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	cutoff := time.Now().UTC().Add(-delta)
-	pkgs, err := handler.scheduler.Poll(cutoff)
-	if err != nil {
-		log.Errorf("error polling for new packages: %v", err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	pkgs, errs := handler.scheduler.Poll(cutoff)
+	if len(errs) > 0 {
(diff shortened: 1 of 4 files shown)


cmd/scheduled-feed/main.go
```diff
@@ -24,11 +24,11 @@ type FeedHandler struct {
...
 	cutoff := time.Now().UTC().Add(-delta)
-	pkgs, err := handler.scheduler.Poll(cutoff)
-	if err != nil {
-		log.Errorf("error polling for new packages: %v", err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+	pkgs, errs := handler.scheduler.Poll(cutoff)
+	if len(errs) > 0 {
+		for _, err := range errs {
+			log.Errorf("error polling for new packages: %v", err)
```
naveensrinivasan, line 30: Why not return http.Error?

naveensrinivasan (approved):
Thanks!

cmd/scheduled-feed/main.go
jordan-wright: I could adjust things further down to return an http error if any feed returned an error, but I wanted to hold off from short circuiting completely here since that would mean that packages we did get wouldn't get sent upstream.

I think it's worth the extra little bit of code at the bottom so I'll add that and merge.

naveensrinivasan: Cool, sounds good. Thanks

PR #60, opened 2021-02-26
Feat - Included docker build and publish to GitHub

naveensrinivasan:
Included docker build and push action.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,52 @@
+name: Docker
+on:
+  push:
+    branches:
+      - main
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+  # Run tests for any PRs.
+  pull_request:
+env:
+  IMAGE_NAME: packagefeeds
+
+jobs:
+  # Push image to GitHub Packages.
+  push:
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


jordan-wright:
Why do we need to push images to GitHub? Right now, we push them to GCR for use in Cloud Run, and otherwise for testing folks can build it locally.

naveensrinivasan:
For others to pull and run the same in their own ENV. That is the reason for doing it.

naveensrinivasan:
> Why do we need to push images to GitHub? Right now, we push them to GCR for use in Cloud Run, and otherwise for testing folks can build it locally.

The easier we make it for someone to try it without pulling the source code, with a docker pull. It is likely they are likely to use it.

Also by pushing to GitHub it shouldn't cause any issues IMO. It also gives us to run docker scan's which is much easier to address when there are security vulnerabilities.

jordan-wright (approved):
I'm not opposed to adding this and it seem relatively straightforward, but in general it's worth noting that it's another thing we'll need to maintain in the future.

As we move forward, I'd gently caution us to be careful about striking the balance between accessibility and being opinionated. Personally, I prefer to hold off on adding new development/deployment models until there's a clear request from the community, rather than adding that complexity in proactively.

I only mention it here since in other projects like Gophish I've been bitten by trying to support every possible deployment case and got overwhelmed with support requests as the edge cases started popping up.

Though it's entirely possible that there already are requests to be able to pull the images down from GitHub and I haven't heard them 😄

PR #61, opened 2021-02-26
Feat - Included docker-scan for the images

naveensrinivasan:
Docker scan for the image.

### DIFF
--- a/.github/workflows/docker-scan.yml
+++ b/.github/workflows/docker-scan.yml
@@ -1,0 +1,32 @@
+name: "docker-scan"
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 1'
+  watch:
+    types: [started]
+
+jobs:
+  Trivy-Scan:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Build an image from Dockerfile
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-scan.yml
```diff
@@ -0,0 +1,32 @@
+name: "docker-scan"
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 1'
```
jordan-wright, line 6: Just a heads up, this appears to run at midnight every Monday - is that what you're looking to do?

.github/workflows/docker-scan.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Build an image from Dockerfile
+        run: |
+          docker build -t docker.pkg.github.com/ossf/package-feeds:${{ github.sha }} .
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@master
```
jordan-wright, line 22: Can we pin this to a version or commit? It's my understanding that using actions from master isn't really recommended, if possible (even though this particular one looks lower risk).

jordan-wright (commented):
This seems reasonable to me! Left a couple of comments for consideration.

PR #65, opened 2021-02-27
Feat - Included distroless container

naveensrinivasan:
Included distroless container to avoid vulnerabilities.
Also included ENV CGO_ENABLED=0  to the docker file to avoid using CGO

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM golang:1.15-buster as builder
 
 # Create and change to the app directory.
 WORKDIR /app
-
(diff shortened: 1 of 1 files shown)


jordan-wright (approved):
Looks good to me! I confirmed the image still builds and runs as expected.

I'm not super worried about vulnerabilities since this is all reached from GCP's internal services, but it's still good for us to set good examples and this certainly doesn't hurt anything.

Thanks for taking the time to knock this one out! Feel free to merge whenever.

PR #68, opened 2021-03-05
Initial NuGet feed support

### LINKED ISSUE #31: .Net/Nuget Feeds
https://www.nuget.org/

FireEater64:
Not sure if there's still interest in this, but this is a rough initial implementation of a nuget feed parser - which I think would close #31. This parser uses the [Catalog](https://docs.microsoft.com/en-us/nuget/api/catalog-resource) endpoint, which provides an append-only feed of nuget package operations performed against a nuget repository (in this case nuget.org).

Parsing the catalog feed is done as follows:

- Resolve the Catalog service endpoint from the [service endpoint](https://api.nuget.org/v3/index.json). This is not strictly necessary (could be hard coded to save a HTTP call), but allows this code to be pointed at any compliant NuGet v3 API
- Fetch the [catalog index document](https://api.nuget.org/v3/catalog0/index.json). This page contains pointers to the leaf catalog nodes.
- Fetch each [catalog leaf document](https://api.nuget.org/v3/catalog0/page12067.json) that has been modified since the cutoff. These documents contain a log of package updates, deletions, reindexings etc.
- For each update newer than the cutoff, fetch details for the package. This is necessary because packages [may be reflowed](https://docs.microsoft.com/en-us/nuget/api/catalog-resource#package-details-catalog-items) into the catalog after initial publication - so we fetch the actual publication date from the leaf detail document

There's definitely some room for optimisation around fetching the leaf pages/package information in parallel, but performance should hopefully be reasonable. Apologies in advance for my rusty golang skills - happy for this to be hacked about with/thrown away!

### DIFF
--- a/feeds/nuget/nuget.go
+++ b/feeds/nuget/nuget.go
@@ -11,8 +11,14 @@ import (
 
 const (
 	FeedName           = "nuget"
-	feedURL            = "https://api.nuget.org/v3/index.json"
 	catalogServiceType = "Catalog/3.0.0"
+)
+
+var (
+	feedURL    = "https://api.nuget.org/v3/index.json"
+	httpClient = http.Client{
+		Timeout: 10 * time.Second,
+	}
 )
 
 type serviceIndex struct {
@@ -44,10 +50,6 @@ type nugetPackageDetails struct {
 	PackageID string    `json:"id"`
 	Version   string    `json:"version"`
 	Created   time.Time `json:"published"`
-}
-
-var httpClient = http.Client{
-	Timeout: 10 * time.Second,
 }
 
 func fetchCatalogService() (*nugetService, error) {
--- a/feeds/nuget/nuget.go
+++ b/feeds/nuget/nuget.go
@@ -15,7 +15,7 @@ const (
 	catalogServiceType = "Catalog/3.0.0"
 )
 
-type nugetServiceIndex struct {
+type serviceIndex struct {
 	Services []*nugetService `json:"resources"`
 }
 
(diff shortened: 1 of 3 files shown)


feeds/nuget/nuget.go
```diff
@@ -0,0 +1,174 @@
...
+type catalogPageLeaf struct {
+	URI            string    `json:"@id"`
+	CatalogCreated time.Time `json:"commitTimeStamp"`
+	Type           string    `json:"@type"`
+}
+
+type nugetPackageDetails struct {
+	PackageID string    `json:"id"`
+	Version   string    `json:"version"`
+	Created   time.Time `json:"published"`
```
FireEater64, line 46: One small uncertainty I had was around whether it would be more appropriate to use the `published` or `created` property of the package details here. Per [the documentation](https://docs.microsoft.com/en-us/nuget/api/catalog-resource#package-details-catalog-items):

> The created timestamp is when the package was first received by the package source, which is typically a short time before the catalog item's commit timestamp.

> The published timestamp is the time when the package was last listed.

My understanding is that if a package is delisted (made private), and then relisted - `published` will be updated, but `created` will not be. By using the `published` property here delisting and relisting a package will trigger another event in the package feed. Is this desirable, or should I be using `created` here instead?

feeds/nuget/nuget.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+var httpClient http.Client = http.Client{
+	Timeout: 10 * time.Second,
+}
+
+func fetchCatalogService() (*nugetService, error) {
+	resp, err := httpClient.Get(feedURL)
+
```
jordan-wright, line 55: As the smallest nit, you could consider removing the spacing between where an error is initialized and where we check it.

Spacing really isn't a big deal to me and is a total nonblocker, but it'll help match the style of the other feeds a bit better.

jordan-wright (changes requested):
This is looking great!

> Apologies in advance for my rusty golang skills

Don't be hard on yourself. Aside from some very minor nits that I've added as comments, this looks perfectly fine to me! And there is definitely interest in supporting any packages we can. 😄 

I should have some time to actually take this for a spin tomorrow, but I wanted to send the first round of feedback ASAP.

Thanks so much for taking the time to send this in. It's very appreciated!

feeds/nuget/nuget.go
jordan-wright: Hmmm.... I don't know that I have a super strong preference. I see pros/cons.

I think keeping it as the published date is fine for now, since I see this as sort of an append-log of packages that we're pushing out. So in that case, it'd be expected to have multiple records that properly indicate that a package was published more than once.

jordan-wright:
LGTM! Thank you for sending this in!

PR #73, opened 2021-03-17
Encapsulate feeds.Package{} struct creation via constructor func

tom--pollard:
This also adds logging output per Package, inline with goproxy feed

### DIFF
--- a/feeds/feed.go
+++ b/feeds/feed.go
@@ -1,6 +1,11 @@
 package feeds
 
-import "time"
+import (
+	"fmt"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
 
 type ScheduledFeed interface {
 	Latest(cutoff time.Time) ([]*Package, error)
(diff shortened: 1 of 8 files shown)


feeds/feed.go
```diff
@@ -12,3 +15,17 @@ type Package struct {
 	CreatedDate time.Time `json:"created_date"`
 	Type        string    `json:"type"`
 }
+
+func NewPackage(created, cutoff time.Time, name, version, feed string) *Package {
+	if created.Before(cutoff) {
+		return nil
+	}
+	log.Println("Processing:", name, version, feed)
```
Qinusty, line 23: Consider using `log.WithFields(...)` here as seen in `cmd/scheduled-feed/main.go`.
```
log.WithFields(log.Fields{
	"name":         pkg.Name,
	"feed":         pkg.Type,
	"created_date": pkg.CreatedDate,
}).Print("sending package upstream")
```

tom--pollard: Sure, also with that in mind it might also be a consideration to globally switch to `logrus` throughout

tom--pollard:
Updated to include nuget & packagist

tom--pollard:
Is this ok to merge, @jordan-wright @dlorenc ? :smiley:

Qinusty (approved):
Looks good @tom--pollard. ~~Only Nit is that I'm starting to think maybe it's cleaner if we handle cutoff in Scheduler.Poll() rather than inside the package creation. This moves responsibility away from the individual feeds and onto the polling service. Happy to open this as a separate issue/PR though.~~

Edit: Although... Some feed APIs may benefit from the cutoff as a query parameter... Let's leave it as is :)

PR #74, opened 2021-03-17
Add configuration options for scheduled-feed binary

Qinusty:
Feeds and publishers have been slightly reworked to support
configuration in place of constant endpoints. In addition to this
it is now possible to support a limited subset of feeds and also multiple
feeds of the same type with different endpoints.

Default behaviour has been preserved through a combination of default configurations
and mutating configurations post-load based on environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,32 @@ infrastructure to study behavior of open source packages and to look for malicio
 We also hope that the components can be used independently, to provide package feeds or runtime
 behavior data for anyone interested.
 
+# Configuration
+
+A YAML configuration file can be provided with the following format:
+
+```
+enabled_feeds:
+- pypi
+- npm
+- goproxy
+- rubygems
+- crates
+
+publisher:
+  type: 'gcp_pubsub'
+  config:
+    url: "gcppubsub://foobar.com"
+
+http_port: 8080
+```
+
+To specify this configuration file, define its path in your environment under the `PACKAGE_FEEDS_CONFIG_PATH` variable.
+
+## Legacy Configuration
+
+Legacy configuration methods are still supported. By default, without a configuration file all feeds will be enabled. The environment variable `OSSMALWARE_TOPIC_URL` can be used to select the GCP pubsub publisher and `PORT` will configure the port for the HTTP server.
+
 # Contributing
 
 If you want to get involved or have ideas you'd like to chat about, we discuss this project in the [OSSF Securing Critical Projects Working Group](https://github.com/ossf/wg-securing-critical-projects) meetings.
--- a/cmd/scheduled-feed/main.go
+++ b/cmd/scheduled-feed/main.go
@@ -10,6 +10,8 @@ import (
 
 	"github.com/ossf/package-feeds/feeds/scheduler"
 	"github.com/ossf/package-feeds/publisher"
+	"github.com/ossf/package-feeds/publisher/gcppubsub"
+	"github.com/ossf/package-feeds/publisher/stdout"
 
 	log "github.com/sirupsen/logrus"
 )
@@ -62,9 +64,9 @@ func main() {
 	var pub publisher.Publisher
 	var err error
 	if pubURL == "" {
-		pub = publisher.NewStdoutPublisher()
+		pub = stdout.New()
 	} else {
-		pub, err = publisher.NewPubSub(context.TODO(), pubURL)
+		pub, err = gcppubsub.New(context.TODO(), pubURL)
 		if err != nil {
 			log.Fatalf("error creating gcp pubsub topic with url %q: %v", pubURL, err)
 		}
--- a/cmd/scheduled-feed/main.go
+++ b/cmd/scheduled-feed/main.go
@@ -7,11 +7,11 @@ import (
 	"net/http"
 	"os"
 	"time"
+	"strings"
 
+	"github.com/ossf/package-feeds/config"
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -15,6 +15,52 @@ infrastructure to study behavior of open source packages and to look for malicio
 We also hope that the components can be used independently, to provide package feeds or runtime
 behavior data for anyone interested.
 
+# Configuration
+
+A configuration file can be provided with the following format:
```
tom--pollard, line 20: Worth explicitly noting that it's yaml?

feeds/crates/crates.go
```diff
@@ -63,8 +72,16 @@ func (feed Feed) Latest(cutoff time.Time) ([]*feeds.Package, error) {
...
 			Version:     pkg.NewestVersion,
-			Type:        FeedName,
+			Type:        feed.GetName(),
 		})
 	}
 	return pkgs, nil
 }
+
+func (feed Feed) GetName() string {
+	parts := []string{FeedType}
```
tom--pollard, line 82: As each feed has to implement this interface & as it stands (afaict?) each is implemented in the same way, is it worth extracting the logic out into a single method?

jordan-wright:
Hey there!

Thanks for sending this in. I'm a bit curious about the use cases here. Typically, I try to avoid making things too configurable since it can make the barrier to entry a bit higher. That is, I try to pick good defaults wherever possible and avoid the amount of things the user has to setup or consider before using the project.

In this case, it looks like the only things that we're making configurable are the feed URLs and the publisher. I can see the merits of making the publisher configurable, but I'm not sure about the feeds. Normally, things like the URLs should be static unless we're doing something like testing, right? And in those cases we can just override the URLs to point to whatever we'd like since we're in the same package scope.

I'd love to hear your thoughts and use cases you have in mind!

Qinusty:
> Hey there!

Hi!

> Thanks for sending this in. I'm a bit curious about the use cases here. Typically, I try to avoid making things too configurable since it can make the barrier to entry a bit higher. That is, I try to pick good defaults wherever possible and avoid the amount of things the user has to setup or consider before using the project.

I agree with this, all of the default behaviour remains the same and the average user could go without any configuration!

> In this case, it looks like the only things that we're making configurable are the feed URLs and the publisher. I can see the merits of making the publisher configurable, but I'm not sure about the feeds. Normally, things like the URLs should be static unless we're doing something like testing, right? And in those cases we can just override the URLs to point to whatever we'd like since we're in the same package scope.

Feed configuration was primarily considered as a means of selecting a subset of feeds to enable for those who are not interested in all feeds. As for URL parameterization, this was exposed as it is currently the only meaningful configuration field for all of these feeds. I'm happy for this to be removed as I have no specific use case for defining a different URL outside of testing.

jordan-wright:
I can definitely see a use case for wanting to enable only a subset of feeds. Perhaps the simplest solution for this would be to expose a `feeds` (or `enabled_feeds`) config parameter that accepts a list of feed names (e.g. `pypi`, `npm`, etc.).

I can see something like that and the publisher config being a good starting point for configuration that isn't too complex for the user.

What do you think?

Qinusty:
> I can definitely see a use case for wanting to enable only a subset of feeds. Perhaps the simplest solution for this would be to expose a `feeds` (or `enabled_feeds`) config parameter that accepts a list of feed names (e.g. `pypi`, `npm`, etc.).

Feeds are now configured by the configuration field "enabled_fields" which takes a list of feed names.

> I can see something like that and the publisher config being a good starting point for configuration that isn't too complex for the user.

Sounds good, I've taken a stab at some changes for this!

README.md
```diff
@@ -21,6 +21,32 @@ infrastructure to study behavior of open source packages and to look for malicio
...
+    url: "gcppubsub://foobar.com"
+
+http_port: 8080
+```
+
+To specify this configuration file, define its path in your environment under the `PACKAGE_FEEDS_CONFIG_PATH` variable.
+
+## Legacy Configuration
+
+Legacy configuration methods are still supported. To configure the application with all feeds enabled, the environment variable `OSSMALWARE_TOPIC_URL` can be used to select the GCP pubsub publisher and `PORT` will configure the port for the HTTP server.
```
tom--pollard, line 48: This reads a little off for me, maybe something like `By default, if no configuration file is provided all feeds will be enabled`

cmd/scheduled-feed/main.go
```diff
@@ -58,30 +60,39 @@ func (handler *FeedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	var appConfig *config.ScheduledFeedConfig
+	if useConfig {
+		appConfig, err = config.FromFile(configPath)
 	} else {
-		pub, err = publisher.NewPubSub(context.TODO(), pubURL)
-		if err != nil {
-			log.Fatalf("error creating gcp pubsub topic with url %q: %v", pubURL, err)
-		}
+		appConfig = config.Default()
```
tom--pollard, line 70: I might be pathing through this wrong, but if no config file is provided, then the env vars for `OSSMALWARE_TOPIC_URL/PORT` are not considered?

Qinusty: Fixed.

README.md
Qinusty: Reworded to clarify this

tom--pollard (approved):

PR #75, opened 2021-03-19
Add Pypi feed testing

Qinusty:
Adds some tests to the pypi feed using httptest as a mock endpoint.

### DIFF
--- a/feeds/pypi/pypi.go
+++ b/feeds/pypi/pypi.go
@@ -11,7 +11,6 @@ import (
 
 const (
 	FeedName = "pypi"
-	baseURL  = "https://pypi.org/rss/updates.xml"
 )
 
 type Response struct {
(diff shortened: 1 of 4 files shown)


feeds/pypi/pypi.go
```diff
@@ -69,11 +68,19 @@ func fetchPackages() ([]*Package, error) {
 	return rssResponse.Packages, nil
 }
 
-type Feed struct{}
+type Feed struct {
```
tom--pollard, line 71: I think the addition of the `Feed` struct to abstract the url is worthy of being a separate commit, but that's only a preference

tom--pollard:
+1 for more testing!

feeds/pypi/pypi.go
Qinusty: Agreed :+1:

Qinusty:
Closing as this will be added to https://github.com/ossf/package-feeds/pull/76

PR #76, opened 2021-03-19
Add test cases for the various feeds

Qinusty:
This PR aims to add tests for all of the feeds currently implemented. These tests should be easily extended to be more exhaustive.

- [x] crates
- [x] rubygems
- [x] goproxy
- [x] npm  
- [x] pypi

### DIFF
--- a/feeds/crates/crates.go
+++ b/feeds/crates/crates.go
@@ -10,7 +10,13 @@ import (
 
 const (
 	FeedName = "crates"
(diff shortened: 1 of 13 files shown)


tom--pollard:
+1 for also adding a constructor for the Feed structs.

Qinusty:
Rebased to latest main, updated to remove URLs from struct and use `var` instead.

tom--pollard (approved):

Issue #80, opened 2021-03-30
Provide and validate against a defined 'Package' json schema


tom--pollard:
This should be covered by a test case, and could be validated against an in-tree schema with https://github.com/xeipuuv/gojsonschema, for example. This should provide a reference for downstream 'subscribers'.

PR #81, opened 2021-03-31
Add initial json schema for package struct, with validation test

tom--pollard:
Resolves https://github.com/ossf/package-feeds/issues/80


package.schema.json
```diff
@@ -0,0 +1,38 @@
...
+        "description": "The package version, formatted respective to the given package type"
+      },
+      "created_date": {
+        "type": "string",
+        "description": "RFC 3339 timestamp representing the package creation date",
+        "format": "date-time"
+      },
+      "type": {
+        "type": "string",
+        "description": "The type of package"
```
phumberdroz, line 30: I am just wondering if this could be more descriptive to describe the origin of the package? Or maybe just an example could help.

Qinusty: I think maintaining a list of valid values here is reasonable, given the mapping to FeedName in the repository

tom--pollard: There's an example of `npm` in the `examples` list, I do agree explicitly stating this maps to the `FeedName` is useful!

package.schema.json
```diff
@@ -0,0 +1,39 @@
...
+      "created_date": {
+        "type": "string",
+        "description": "RFC 3339 timestamp representing the package creation date",
+        "format": "date-time"
+      },
+      "type": {
+        "type": "string",
+        "description": "The type of package, this being the `FeedName` of the given package feed"
+      },
+      "schema_ver": {
```
pierre-ernst, line 32: instead of `name`, `version` and `type`, I would love to see support for [packageURL](https://github.com/package-url/purl-spec)

package.schema.json
```diff
@@ -0,0 +1,39 @@
...
+    "type": "object",
+    "properties": {
+      "name": {
+        "type": "string",
+        "description": "The name of the package"
+      },
+      "version": {
+        "type": "string",
+        "description": "The package version, formatted respective to the given package type"
+      },
```
pierre-ernst, line 22: On one hand, each package maintainer is free to use whichever versioning schema they choose fit, so free-text should definitely be accepted. But package versions complying with [semantic versioning](https://semver.org/) makes downstream compute/lookup much easier. As such, it would be awesome to add a flag mentioning that the version string can be parsed with semver libraries.

package.schema.json
tom--pollard: Sure, with versioning of the schema it should allow for extra metadata(such as packageURL, for example) to be added and optionally consumed by subscribers, whilst not breaking compatibility

package.schema.json
Qinusty: I would prefer to avoid having a flag to indicate semver compatible version strings as this can be checked on the consumer side just as easily. Unfortunately semver usage is down to the maintainers of individual packages as you've mentioned, so we can't even define this on a per-feed basis.

tom--pollard:
I've dropped the WIP tag as I think this is in a state where it's reviewable, all be it with the open questions in the description

package.schema.json
```diff
@@ -0,0 +1,39 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://github.com/ossf/package-feeds/blob/main/package.schema.json",
+    "title": "Package Schema Version 1.0",
+    "description": "The package representation as outputted by a ScheduledFeed",
+    "examples": [
```
Qinusty, line 6: This seems strange as a list of values for the potential fields? Is this supposed to be a list of objects with key/values?

Qinusty (commented):
Some minor comments/questions

package.schema.json
```diff
@@ -0,0 +1,39 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
```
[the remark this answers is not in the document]
tom--pollard: Apparently not, the naming is somewhat confusing https://json-schema.org/understanding-json-schema/reference/schema.html but is required by the validator

package.schema.json
tom--pollard: I went off https://json-schema.org/understanding-json-schema/reference/generic.html#annotations for this, which doesn't specify it it should be a list of tuples. I've not got another reference point/real example to hand though

feeds/feed.go
```diff
@@ -7,15 +7,19 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const SchemaVer = "1.0"
```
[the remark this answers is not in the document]
tom--pollard: Agreed, good catch

feeds/feed_test.go
```diff
@@ -0,0 +1,51 @@
...
+	validPackage := gojsonschema.NewGoLoader(dummy_package)
+
+	result, err := gojsonschema.Validate(schemaLoader, validPackage)
+
+	if err != nil {
+		panic(err.Error())
+	}
+
+	if result.Valid() != true {
+		out := fmt.Sprintf("The Package json is not valid against the current schema. see errors :\n")
```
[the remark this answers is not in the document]
tom--pollard: True, thanks!

package.schema.json
Qinusty: Ah it seems like this `examples` field belongs in each of the field objects? This would make more sense for a list of examples. Works perfectly for our `type` value too!

tom--pollard: Pushed an update to have an examples annotation on a per property basis, and also an extra test for tighter control on additional properties. How does it look now @Qinusty ?

feeds/feed_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/xeipuuv/gojsonschema"
+)
+
+const schema_path = "../package.schema.json"
+
+type ExtendPackage struct {
```
Qinusty, line 13: This can be `extendPackage` as it can remain private in scope.

feeds/feed_test.go
```diff
@@ -0,0 +1,72 @@
...
+	*Package
+	NonConformingField string `json:"non_conforming_field"`
+}
+
+func TestValidation(t *testing.T) {
```
Qinusty, line 18: Can we break this out into two tests? One for valid and one for invalid? This makes it clearer in the test logs when one of them fails.

feeds/feed_test.go
tom--pollard: True, thanks!

feeds/feed_test.go
tom--pollard: Yep makes sense, done

Qinusty (approved):
LGTM!

PR #84, opened 2021-04-13
Add Kafka publisher support

Qinusty:
This adds support for kafka pubsub as a publisher implementation.


publisher/kafkapubsub/kafkapubsub.go
```diff
@@ -0,0 +1,47 @@
+package kafkapubsub
+
+import (
+	"context"
+
+	"gocloud.dev/pubsub"
+	"gocloud.dev/pubsub/kafkapubsub"
```
tom--pollard, line 7: Can we blank import this, to match the GCP driver implementation?

Qinusty: This is why we are using kafkapubsub.OpenTopic(), we are aiming to hide the implementation usage of gocloud within the publisher and avoid the ENV var.

tom--pollard: I'm indifferent about this, so happy as is

tom--pollard (approved):

Issue #85, opened 2021-04-13
Enable internally driven scheduling


tom--pollard:
As it stands the current deployment of polling uses a HTTP server to await requests from a cloud scheduler job which will poll all of the configured feeds and publish the packages appropriately. This requires that the internal `delta` constant matches the external frequency of the scheduler job, as well as relying on extra infrastructure.

It is desirable for this polling to be configured (using the recently added configuration file  https://github.com/ossf/package-feeds/pull/74) to use an internal timer and potentially schedule feed polling based on individual timings. This could be done by keep the existing HTTP server model, and using an internal `cron` style job such as https://github.com/robfig/cron to drive requests to localhost. Or, it may be desirable to not run a server at all in this mode of operation. `daemon` is an often used term for this usecase, however that does entail additional control functionality which is potentially out of scope.

Qinusty:
I think that leaving the option open for polling via HTTP request is ideal, but for many use cases this isn't required and simply scheduling the goroutines for polling on intervals would allow for more fine grained control and potentially polling different feeds at different rates. 

Whilst polling faster won't necessarily solve https://github.com/ossf/package-feeds/issues/83 it definitely makes sense to allow feeds to be polled at different intervals depending on the exact API requirements.

tom--pollard:
@Qinusty feed specific poll rates is an interesting thought, would you expect this to be an extension of the enabled feeds config? Say, default to global `cutoff_delta` value for all feeds, unless a value is specified at a feed level at which a separate cron can be func can be generated. Having variable polling rates however would again highlight the need to change the `cutoff` logic from a static delta to that of a value based on the last time polled, potentially on a per feed basis.

tom--pollard:
I think with https://github.com/ossf/package-feeds/pull/122 merged we can now consider this issue complete.

PR #86, opened 2021-04-13
Extract cutoff delta to configuration file

tom--pollard:
This keeps the default behaviour the same, whilst also opening up some needed configurability for https://github.com/ossf/package-feeds/issues/85

### DIFF
--- a/config/structs.go
+++ b/config/structs.go
@@ -4,6 +4,7 @@ type ScheduledFeedConfig struct {
 	PubConfig    PublisherConfig `yaml:"publisher"`
 	EnabledFeeds []string        `yaml:"enabled_feeds"`
 	HttpPort     int             `yaml:"http_port,omitempty"`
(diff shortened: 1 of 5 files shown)


config/structs.go
```diff
@@ -4,6 +4,7 @@ type ScheduledFeedConfig struct {
 	PubConfig    PublisherConfig `yaml:"publisher"`
 	EnabledFeeds []string        `yaml:"enabled_feeds"`
 	HttpPort     int             `yaml:"http_port,omitempty"`
+	CutoffDelta  string          `yaml:"time_delta"`
```
Qinusty, line 7: time_delta and cutoff_delta are used in different parts of this PR, these will need to be updated to be consistent.

tom--pollard: Ah sorry, missed replacing that one

Qinusty (approved):
Looks good to me :+1:

Issue #83, opened 2021-04-09
Addressing lossy feeds


Qinusty:
The firehose from pypi `Latest()` can occasionally miss package updates due to a limitation on the 50 most recent package updates in the current pypi `updates.xml` rss feed. In some use cases this is acceptable but for certain deployments it is critical that certain package updates are not missed.

I propose two solutions for discussion:

- Extend the Feeds API to support querying a subset of packages which is optional for a given feed to implement. This allows feeds like Pypi to take advantage of these package specific APIs to ensure that updates are not missed at the cost of additional requests to the API. This will allow other implementations to avoid loss of package updates also.
- Augment the existing Pypi feed to be configurable with this `critical` package list and check these manually alongside the standard `updates.xml` firehouse.

For Pypi, I have a WIP implementation of the necessary API calls in a custom feed at https://github.com/Qinusty/package-feeds/commit/deeeec7f606272c2674f23d18f9b96b130c2409d

I'm interested in hearing peoples thoughts and use cases for this.

tom--pollard:
I think pypi is an excellent example here, even if the default poll rate (5 minutes) was tightened there's still scope to missing updates due to the 50 'snapshot' limitation on the general  `updates.xml` rss feed.  Obviously that scope still exists when hitting the specific package release endpoints, however much more unlikely when restricted to a single package.

georgevanburgh:
Does this apply to some of the other feeds as well? The npm feed appears to use a rss feed of limited length

https://github.com/ossf/package-feeds/blob/97a1ef475bfc6d22e716a3a9405fa69c122ac39c/feeds/npm/npm.go#L15

As does the cargo feed

https://github.com/ossf/package-feeds/blob/97a1ef475bfc6d22e716a3a9405fa69c122ac39c/feeds/crates/crates.go#L13

I'm not sure what the rules are for these endpoints (whether they expire from the feed based on time, or number of packages, or both) - but it feels like some of them might have the same issues.

Another thing that might come up as part of this - currently each feed is parsed in its own goroutine, meaning feeds are processed in parallel - but the processing within each feed is sequential. Some of the existing feeds (npm, nuget) make a request to the registry for each package they process - and do so sequentially, which makes processing these feeds significantly slower than feeds like cargo, where only a single request is required per scan. It looks like the proof-of-concept linked above introduces parallelism within the `pypi_critical` feed itself, which is really cool - and it might also be worth extending this parallelism to the other feeds mentioned above.

[...]

PR #87, opened 2021-04-13
WIP: Add pypi_critical feed

Qinusty:
This implementation is subject to change and serves as an illustration of one way we could approach the problem detailed in #83. Adding this to a new feed is a simple way of handling this for one use case (pypi) but if the problem exists in many feeds it may be better to adapt the `ScheduledFeed` interface to provide some functionality to query APIs for specific 'critical' packages where loss of updates is simply not an option.

### DIFF
--- a/config/structs.go
+++ b/config/structs.go
@@ -1,12 +1,13 @@
 package config
 
 type ScheduledFeedConfig struct {
-	PubConfig    PublisherConfig `yaml:"publisher"`
-	EnabledFeeds []string        `yaml:"enabled_feeds"`
-	HttpPort     int             `yaml:"http_port,omitempty"`
+	PubConfig    TypeConfigPair   `yaml:"publisher"`
+	EnabledFeeds []string         `yaml:"enabled_feeds"`
+	Feeds        []TypeConfigPair `yaml:"feeds"`
+	HttpPort     int              `yaml:"http_port,omitempty"`
 }
 
-type PublisherConfig struct {
+type TypeConfigPair struct {
 	Type   string      `mapstructure:"type"`
 	Config interface{} `mapstructure:"config"`
 }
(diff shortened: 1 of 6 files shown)


config/structs.go
```diff
@@ -1,12 +1,13 @@
 package config
 
 type ScheduledFeedConfig struct {
-	PubConfig    PublisherConfig `yaml:"publisher"`
-	EnabledFeeds []string        `yaml:"enabled_feeds"`
-	HttpPort     int             `yaml:"http_port,omitempty"`
+	PubConfig    TypeConfigPair   `yaml:"publisher"`
+	EnabledFeeds []string         `yaml:"enabled_feeds"`
+	Feeds        []TypeConfigPair `yaml:"feeds"`
```
tom--pollard, line 6: Struggling to visualise how the config file would look for this. As in, how would you know to list 'Feeds that require type/config' under `feeds`, and generic feeds under `enabled_feeds`? Maybe it's just a case of a using a different fieldname

Qinusty: The idea here is perhaps preliminary but maintaining both EnabledFeeds and Feeds to allow configuration of these feeds (specifically pypi_critical in this case).

An example config file for this might look as follows:

```yaml
publisher:
  type: stdout

enabled_feeds:
  - pypi
  - rubygems
  - pypi_critical

feeds:
  - type: pypi_critical
    config:
      packages:
        - foopackage
        - barpackage
        - bazpackage
```

The whole type/config pair is something I'm keen to work around if there's a better option. The problem being the `packages` field containing a list of ScheduledFeed configurations but you can't unmarshal these interfaces without a hint towards the struct i.e. `type`

Qinusty:
This PR will be reworked to introduce this functionality into the existing pypi feed as per https://github.com/ossf/package-feeds/issues/83#issuecomment-820392580

Qinusty:
Closing this PR to follow the direction of issue https://github.com/ossf/package-feeds/issues/83 and integrate this feature into pypi directly with configuration options.

PR #88, opened 2021-04-15
Add basic cron based 'timer' mode

tom--pollard:
This leverages the `cutoff_delta` config, adding the option to set an internal `cron` like scheduler for triggering the polling of feeds. The current implementation somewhat mimics the current trigger which is driven by requests from an external scheduled [service](https://github.com/ossf/package-feeds/blob/main/terraform/scheduler/main.tf). 

As mentioned in #85 this is one approach that uses a static frequency, this could be expanded to be dynamic (storing an internal value for `time_of_last_poll`, to use for the `cuttoff` threshold instead potentially on a per feed basis). Further to that is the consideration of if a httpserver should even be launched in this `standalone` mode, or even a hybrid approach. These can all be considered in followup PRs

### DIFF
--- a/cmd/scheduled-feed/main.go
+++ b/cmd/scheduled-feed/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ossf/package-feeds/feeds/scheduler"
 	"github.com/ossf/package-feeds/publisher"
 
+	"github.com/robfig/cron"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -97,8 +98,27 @@ func main() {
 		pub:       pub,
 		delta:     delta,
 	}
+
+	if appConfig.Timer {
+		cronjob := cron.New()
+		crontab := fmt.Sprintf("@every %s", delta.String())
+		log.Printf("Running a timer %s", crontab)
+		cronjob.AddFunc(crontab, func() { cronrequest(appConfig.HttpPort) })
+		cronjob.Start()
+	}
+
 	http.Handle("/", handler)
 	if err := http.ListenAndServe(fmt.Sprintf(":%v", appConfig.HttpPort), nil); err != nil {
 		log.Fatal(err)
+	}
+}
+
+func cronrequest(port int) {
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
+	_, err := client.Get(fmt.Sprintf("http://localhost:%v", port))
+	if err != nil {
+		log.Printf("http request failed: %v", err)
 	}
 }
--- a/README.md
+++ b/README.md
@@ -41,9 +41,12 @@ publisher:
 http_port: 8080
 
 cutoff_delta: 5m
+
+timer: false
 ```
 
-`cutoff_delta` string formatted for [duration parser](https://golang.org/pkg/time/#ParseDuration). To specify this configuration file, define its path in your environment under the `PACKAGE_FEEDS_CONFIG_PATH` variable.
+`cutoff_delta` string formatted for [duration parser](https://golang.org/pkg/time/#ParseDuration).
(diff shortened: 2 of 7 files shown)


cmd/scheduled-feed/main.go
```diff
@@ -97,8 +98,25 @@ func main() {
...
 	http.Handle("/", handler)
 	if err := http.ListenAndServe(fmt.Sprintf(":%v", appConfig.HttpPort), nil); err != nil {
 		log.Fatal(err)
 	}
 }
+
+func cronrequest(port int) {
+	resp, err := http.Get(fmt.Sprintf("http://localhost:%v", port))
+	if err != nil {
+		log.Printf("http request failed")
```
naveensrinivasan, line 119: I would add the error message to the log

cmd/scheduled-feed/main.go
```diff
@@ -97,8 +98,25 @@ func main() {
...
+	}
+
 	http.Handle("/", handler)
 	if err := http.ListenAndServe(fmt.Sprintf(":%v", appConfig.HttpPort), nil); err != nil {
 		log.Fatal(err)
 	}
 }
+
+func cronrequest(port int) {
+	resp, err := http.Get(fmt.Sprintf("http://localhost:%v", port))
```
naveensrinivasan, line 117: It would good to have timeout on HTTP get.

cmd/scheduled-feed/main.go
```diff
@@ -97,8 +98,25 @@ func main() {
 		pub:       pub,
 		delta:     delta,
 	}
+
+	if appConfig.Timer {
+		cronjob := cron.New()
+		crontab := fmt.Sprintf("@every %s", delta.String())
```
naveensrinivasan, line 104: Could avoid declaring the variable `crontab` and include the `Sprintf` in the log.

naveensrinivasan (commented):
👍

cmd/scheduled-feed/main.go
tom--pollard: The var is also used as a parameter the cron `AddFunc()` call else I would agree that it's unnecessary to declare it just for the log

cmd/scheduled-feed/main.go
tom--pollard: This raises another question, should this be a `log.Fatal()`, as it stands if a scheduled call was to fail (or be delayed waiting on a timeout) then upto 5 minutes of events could be naively missed per cron.

cmd/scheduled-feed/main.go
tom--pollard: Added a timeout inline with the codebase

cmd/scheduled-feed/main.go
tom--pollard: `err` added to the log call, thanks

cmd/scheduled-feed/main.go
```diff
@@ -97,8 +98,28 @@ func main() {
...
+
+func cronrequest(port int) {
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
+	resp, err := client.Get(fmt.Sprintf("http://localhost:%v", port))
+	if err != nil {
+		log.Printf("http request failed: %v", err)
+	}
+	defer resp.Body.Close()
```
Qinusty, line 124: defer isn't required here as we're not going on to do anything.

Qinusty (commented):
Mostly looks good to me, minor comment.

cmd/scheduled-feed/main.go
tom--pollard: Agreed, thanks

Qinusty (approved):
LGTM!

PR #89, opened 2021-04-16
Add event framework (Log Lossy Feed events)

Qinusty:
This PR introduces an event framework for driving defined system events
through a sink interface. A logrus logging sink is implemented
initially for logging event locally but event could also be
dispatched to alternative metrics/logging mechanisms.

The initial event which is added for this is to log and
identify potential loss of data between polling intervals.
This lossy feed problem is detailed in issue #83 and whilst
this doesn't fix the problem, it can help to identify and monitor
occurrences of this issue within a long running production environment.

Lossy feed event checking and dispatch is added as a utility in the feeds package 
and is implemented within the currently supported feeds which have the potential to be lossy 
(crates, npm, pypi, rubygems).

### DIFF
--- a/config/scheduledfeed.go
+++ b/config/scheduledfeed.go
@@ -88,22 +88,26 @@ func AddTo(ls *[]int, value int) {
 func (config *ScheduledFeedConfig) GetScheduledFeeds() (map[string]feeds.ScheduledFeed, error) {
 	var err error
 	scheduledFeeds := map[string]feeds.ScheduledFeed{}
+	eventHandler, err := config.GetEventHandler()
+	if err != nil {
+		return nil, err
+	}
 	for _, entry := range config.EnabledFeeds {
 		switch entry {
 		case crates.FeedName:
-			scheduledFeeds[entry] = crates.Feed{}
+			scheduledFeeds[entry] = crates.New(eventHandler)
 		case goproxy.FeedName:
 			scheduledFeeds[entry] = goproxy.Feed{}
 		case npm.FeedName:
-			scheduledFeeds[entry] = npm.Feed{}
+			scheduledFeeds[entry] = npm.New(eventHandler)
 		case nuget.FeedName:
 			scheduledFeeds[entry] = nuget.Feed{}
 		case pypi.FeedName:
-			scheduledFeeds[entry] = pypi.Feed{}
+			scheduledFeeds[entry] = pypi.New(eventHandler)
 		case packagist.FeedName:
 			scheduledFeeds[entry] = packagist.Feed{}
 		case rubygems.FeedName:
-			scheduledFeeds[entry] = rubygems.Feed{}
+			scheduledFeeds[entry] = rubygems.New(eventHandler)
 		default:
 			err = fmt.Errorf("%w : %v", errUnknownFeed, entry)
 		}
--- a/config/scheduledfeed.go
+++ b/config/scheduledfeed.go
@@ -11,8 +11,10 @@ import (
 	"strconv"
 
 	"github.com/mitchellh/mapstructure"
+	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 
+	"github.com/ossf/package-feeds/events"
 	"github.com/ossf/package-feeds/feeds"
 	"github.com/ossf/package-feeds/feeds/crates"
 	"github.com/ossf/package-feeds/feeds/goproxy"
@@ -28,8 +30,9 @@ import (
 )
 
 var (
-	errUnknownFeed = errors.New("unknown feed type")
-	errUnknownPub  = errors.New("unknown publisher type")
+	errUnknownFeed     = errors.New("unknown feed type")
+	errUnknownPub      = errors.New("unknown publisher type")
+	errUnknownSinkType = errors.New("unknown sink type")
 )
 
 // Loads a ScheduledFeedConfig struct from a yaml config file.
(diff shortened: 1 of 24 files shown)


events/handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (e LossyFeedEvent) GetComponent() string {
+	return FEEDS_COMPONENT
+}
+
+func (e LossyFeedEvent) GetType() string {
+	return LOSSY_FEED_EVENT
+}
+
+func (e LossyFeedEvent) GetMessage() string {
```
tom--pollard, line 122: I wonder if for this specific event we'd want to expand the struct to store some of the date/time info which could be added to the log. This could be a useful datapoint for external investigation, for instance a given timeframe in which we can gather that no events were recorded? Obviously this could just be a false positive and an event may have genuinely not occurred during the delta , but it's at least a data point.

feeds/scheduler/scheduler.go
```diff
@@ -51,6 +63,29 @@ func (s *Scheduler) Poll(cutoff time.Time) ([]*feeds.Package, []error) {
...
+		}
+		s.previousPollResults[result.name] = result
 	}
 	return packages, errs
 }
+
+// Checks whether there is an overlap in package creation date between a result
+// and a previous result. This assumes that pollResult.packages is sorted by
+// CreatedDate in order of most recent first.
+func (r pollResult) isOverlappingWith(previousResult pollResult) bool {
```
tom--pollard, line 84: With the current static polling rate and how it is then used to create a cutoff point, isn't the current expectation here that there will be no overlaps (as they've already been filtered by the cutoff when the Package structs are initialised?) between each scheduled poll, assuming singular instance?

Qinusty: Indeed, I think we will need to move the cutoff calculation outside of the feeds to allow for these feeds to be stored and compared.

The filter of packages with cutoff can be done prior to the return of `Poll()`

feeds/feed.go
```diff
@@ -18,20 +15,21 @@ type Package struct {
 	Type        string    `json:"type"`
 }
 
-func NewPackage(created, cutoff time.Time, name, version, feed string) (*Package, error) {
-	if created.Before(cutoff) {
-		return nil, fmt.Errorf("package was created before cutoff time: %s", cutoff.String())
-	}
-	log.WithFields(log.Fields{
```
tom--pollard, line 25: Should this log be re-added to the ApplyCuttoff func?

feeds/scheduler/scheduler.go
tom--pollard: This comment is null now, as it's applied pre-cutoff filtering

events/handler.go
Qinusty: I'm not sure whether we want this in the event structure itself but perhaps in the sink implementations, kafka for example comes with timestamping built in so we would be providing multiple timestamp values. Logrus could also be configured with a full timestamp field in the log message.

feeds/feed.go
Qinusty: This is added to `scheduler.go` as part of `Poll()`

tom--pollard: I don't have a strong preference but it is a change in behaviour as it was only logged after the cutoff was applied

Qinusty: Cutoff is applied inside the feeds. Only post-cutoff packages are logged. The difference in behaviour is the timing of the logging, logging occurs after all packages have been fetched for a given feed instead of during processing.

tom--pollard: Yeh fair enough, no user-facing difference here :+1:

events/handler.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (h Handler) GetFilter() Filter {
+	return h.eventFilter
+}
+
+// Checks whether an event should be dispatched under the configured
+// filter options.
+// Options are applied as follows:
+// 	- disabled event types are always disabled.
```
tom--pollard, line 86: nit: extra space

config/scheduledfeed.go
```diff
@@ -105,6 +111,33 @@ func (sConfig *ScheduledFeedConfig) GetScheduledFeeds() (map[string]feeds.Schedu
...
+
+func (ec EventsConfig) ToEventHandler() (*events.Handler, error) {
+	var sink events.Sink
+	switch ec.Sink {
+	case events.LOGGING_SINK_TYPE:
+		sink = events.NewLoggingEventSink(log.New())
+	default:
+		return nil, fmt.Errorf(
+			"unknown sink type '%v' provided to events configuration",
+			ec.Sink,
```
tom--pollard, line 135: Could be nice to provide the available sink types here in the error message?

tom--pollard:
As this is essentially opening up a new data type that can be put 'on the wire', (including the scope to have pubsub `sinks` in a similar fashion to the existing package publishers) does it make sense to have a defined expectation on what constitutes an 'event'?. 

The event introduced here exercises all of the methods provided by the interface (component, type, message) when generating the logrus output, Could `DispatchEvent()` potentially pass a validated struct/schema constructed from the interface methods to the event handlers `AddEvent()` instead of the event itself?

events/lossy_feed_event.go
```diff
@@ -0,0 +1,23 @@
+package events
+
+import (
+	"fmt"
+)
+
+type LossyFeedEvent struct {
+	Feed string
+
+	Event
```
tom--pollard, line 10: Not clear on what `Event` is for here?

Qinusty:
> As this is essentially opening up a new data type that can be put 'on the wire', (including the scope to have pubsub `sinks` in a similar fashion to the existing package publishers) does it make sense to have a defined expectation on what constitutes an 'event'?.

I think that the definition of an `event` can remain quite flexible, the exact reasoning behind a new event can be discussed during the implementation and review of specific event additions. As for how verbose this becomes, we can review this at a later date but the sinks can dispatch events in a filterable manner e.g. to specific type topics for a kafka sink (for example).

> The event introduced here exercises all of the methods provided by the interface (component, type, message) when generating the logrus output. Could `DispatchEvent()` potentially pass a struct (validated with a schema for expected fields, format etc as with packages) constructed from the interface methods to the event handlers `AddEvent()` instead of the event itself?

Yeah I think this might need some tweaking to have an easily method of encoding data for transmission. We might want some form of standardised event payload which the individual Event implementations can populate. The current method of having the sink switch based on the event type isn't very extendable for many sinks and many events.

events/handler.go
tom--pollard: I think that's subtly different to what I'm suggesting there. These would be separate timestamps pulled from the packages themselves (most recent package from previous results, oldest package from current results) as oppose to the timestamp at which the event itself was generated. This could be represented as a formatted string based field to avoid potential confusion

config/scheduledfeed.go
Qinusty: This would be easy to do now but would add an additional burden to adding sinks which could easily be missed. I think that adding documentation for the currently supported sinks/publishers/feeds etc is something that should be done for supporting user configuration. https://github.com/ossf/package-feeds/issues/97 is open to track this.

Qinusty:
Extending the event framework for event serialization for future sinks is being tracked as part of issue https://github.com/ossf/package-feeds/issues/96.

config/scheduledfeed.go
tom--pollard: Agreed

events/handler.go
tom--pollard: As events are not formalised I think any extension to this or future event types can be proposed as followups, just to clarify that this is just a suggestion :smile:

config/scheduledfeed.go
tom--pollard: Great! Until #97 is resolved it may be worth just adding a small snippet to the example config in the `README` as part of this PR, highlighting it optionally enabled (not blocking on this however).

tom--pollard (approved):
Other than my open 'nit' comment this all looks good to me, feel free to resolve the README suggestion comment

Issue #77, opened 2021-03-22
Standardise linting/CI checks across packages


Qinusty:
Currently `feeds/npm` and `feeds/goproxy` have makefiles in place to run strict linting checks, linting/formatting strictness should probably be made consistent throughout the project.

PR #91, opened 2021-04-20
Expand ci build & verify across all feeds with workflow matrix

tom--pollard:
This looks to work towards  #77 

As lint/format commits are somewhat noisy,  it may make sense to apply build/verify to the entire codebase (not just the feeds) in a single PR instead of gradually. Also there may be a preference for having the lint commit before instead of after the commit in which the feature is enabled in CI, as such each commit **passes**.

 Using workflow matrices allows the parallelisation across feed packages whilst reducing the amount of repetition in the source files, or reliance on Make Recursion.


tom--pollard:
This is superseed by #92

PR #94, opened 2021-04-26
Use lastPoll time as cutoff on subsequent polls

tom--pollard:
This is additional work towards #85 

Using the last poll time as a cutoff marker on subsequent polls should help with robustness against missed events, for instance if a previously scheduled poll event was missed or delayed. In reverse an unscheduled poll event could be triggered in an interval, losing the assumed synchronisation of the static cutoff mechanism.  If the configuration surface is extended to allow variable polling rates (per feed, for example) then this could simply be extended to a `map`, or similar.

### DIFF
--- a/cmd/scheduled-feed/main.go
+++ b/cmd/scheduled-feed/main.go
@@ -21,11 +21,13 @@ import (
 type FeedHandler struct {
 	scheduler *scheduler.Scheduler
 	pub       publisher.Publisher
-	delta     time.Duration
+	pollRate  time.Duration
+	lastPoll  time.Time
 }
 
 func (handler *FeedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	cutoff := time.Now().UTC().Add(-handler.delta)
+	cutoff := handler.getCutoff()
+	handler.lastPoll = time.Now().UTC()
 	pkgs, errs := handler.scheduler.Poll(cutoff)
 	if len(errs) > 0 {
 		for _, err := range errs {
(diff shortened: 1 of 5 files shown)


cmd/scheduled-feed/main.go
```diff
@@ -22,10 +22,17 @@ type FeedHandler struct {
 	scheduler *scheduler.Scheduler
 	pub       publisher.Publisher
 	delta     time.Duration
+	lastPoll  time.Time
 }
 
 func (handler *FeedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
Qinusty, line 28: This function is getting quite loaded with logic, we should break the cutoff calculation out to start to simplify this.

cmd/scheduled-feed/main.go
```diff
@@ -22,10 +22,17 @@ type FeedHandler struct {
...
-	cutoff := time.Now().UTC().Add(-handler.delta)
+	var cutoff time.Time
+	if handler.lastPoll.IsZero() {
+		handler.lastPoll = time.Now().UTC()
```
Qinusty, line 31: Will this only happen once during first poll due to `IsZero()`

tom--pollard: Should do, as it stands

Qinusty: We want `lastPoll` to be set every time we poll though right?

tom--pollard: Yep.... should be rectified now

cmd/scheduled-feed/main.go
tom--pollard: Sure I don't disagree there, I've broke this out into a value receiver func

Qinusty (approved):
Looks good to me :+1:

tom--pollard:
As mentioned in the body of the PR, this does introduce a potential change to the default behaviour. As such it could be expanded to be enabled behind a config option.

tom--pollard:
I've updated the PR to change `cutoff_delta` to `poll_rate` to better reflect how its used with this change

Issue #95, opened 2021-04-26
Tweak http response handling in `cronrequest()`


tom--pollard:
As `ServeHTTP` doesn't write a response until all packages from a current 'poll' session have been processed, the current timeout of `10s` on the cron `GET` request times-out waiting for the respons headers.

```
level=info msg="http request failed: Get \"http://localhost:8080\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

The error is non-fatally handled, as such this isn't a critical issue however it obviously should be handled more gracefully, especially when writing to the log.

tom--pollard:
The default behaviour of having no timeout could be suited to this usecase, however this is risky when running the requests in goroutines (via the cron package) as they could be left waiting to exit if the localhost httpserver is down for whatever reason

PR #98, opened 2021-04-29
Poll directly when running internal timer requests

tom--pollard:
~~This is currently based on https://github.com/ossf/package-feeds/pull/94~~

This should resolve https://github.com/ossf/package-feeds/issues/95 by not relying on http requests to the localserver, and is extra progress towards https://github.com/ossf/package-feeds/issues/85

### DIFF
--- a/cmd/scheduled-feed/main.go
+++ b/cmd/scheduled-feed/main.go
@@ -134,7 +134,7 @@ func main() {
 		cronjob := cron.New()
 		crontab := fmt.Sprintf("@every %s", pollRate.String())
 		log.Printf("Running a timer %s", crontab)
-		err := cronjob.AddFunc(crontab, func() { cronRequest(appConfig.HTTPPort) })
+		err := cronjob.AddFunc(crontab, func() { cronPoll(handler) })
 		if err != nil {
 			log.Fatal(err)
 		}
@@ -147,14 +147,16 @@ func main() {
 	}
 }
 
-func cronRequest(port int) {
-	client := &http.Client{
-		Timeout: 10 * time.Second,
-	}
-	resp, err := client.Get(fmt.Sprintf("http://localhost:%v", port))
+func cronPoll(handler *FeedHandler) {
+	pkgs, pollErrors := handler.pollFeeds()
+	processedPackages, err := handler.publishPackages(pkgs)
 	if err != nil {
-		log.Printf("http request failed: %v", err)
+		log.Errorf(err.Error())
 		return
 	}
-	resp.Body.Close()
+	if pollErrors {
+		// pollFeeds already logs with ErrorF.
+		return
+	}
+	log.Printf("%d packages processed", processedPackages)
 }
--- a/cmd/scheduled-feed/main.go
+++ b/cmd/scheduled-feed/main.go
@@ -13,6 +13,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/ossf/package-feeds/config"
+	"github.com/ossf/package-feeds/feeds"
 	"github.com/ossf/package-feeds/feeds/scheduler"
 	"github.com/ossf/package-feeds/publisher"
 )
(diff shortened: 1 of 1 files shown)


cmd/scheduled-feed/main.go
```diff
@@ -131,14 +147,16 @@ func main() {
 	}
 }
 
-func cronRequest(port int) {
-	client := &http.Client{
-		Timeout: 10 * time.Second,
-	}
-	resp, err := client.Get(fmt.Sprintf("http://localhost:%v", port))
+func cronRequest(handler *FeedHandler) {
```
Qinusty, line 150: This can be renamed now to reflect the changes in this PR as it no longer makes the http request.

cmd/scheduled-feed/main.go
```diff
@@ -45,33 +79,15 @@ func (handler *FeedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		b, err := json.Marshal(pkg)
 		if err != nil {
 			log.Printf("error marshaling package: %#v", pkg)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
+			return processed, err
```
Qinusty, line 82: I think that failing to marshal a pkg should be a soft failure, we can continue to publish the next package, with processed++ only happening on a successful publish. 

Obviously a failed publish isn't worth retrying as it is likely a problem with the publisher configuration or endpoint.

cmd/scheduled-feed/main.go
```diff
@@ -26,17 +27,50 @@ type FeedHandler struct {
...
+
+func (handler *FeedHandler) pollFeeds() ([]*feeds.Package, bool) {
 	cutoff := handler.getCutoff()
 	handler.lastPoll = time.Now().UTC()
 	pkgs, errs := handler.scheduler.Poll(cutoff)
+	errors := false
 	if len(errs) > 0 {
+		errors = true
 		for _, err := range errs {
 			log.Errorf("error polling for new packages: %v", err)
```
Qinusty, line 65: Whilst this isn't part of this PR, this seems duplicative as the scheduler already has `logger.WithError(result.err).Error("error fetching packages")` as it appends to the errors return value.

cmd/scheduled-feed/main.go
```diff
@@ -131,14 +147,16 @@ func main() {
 	}
 }
 
-func cronRequest(port int) {
-	client := &http.Client{
-		Timeout: 10 * time.Second,
-	}
-	resp, err := client.Get(fmt.Sprintf("http://localhost:%v", port))
+func cronRequest(handler *FeedHandler) {
+	pkgs, pollErrors := handler.pollFeeds()
```
Qinusty, line 151: Does the linter complain if pollErrors is left as `_`? Given it isn't an error type I'm not sure of the linter behaviour

Qinusty (commented):
Some thoughts, looks good though!

cmd/scheduled-feed/main.go
tom--pollard: I don't disagree, but this is the original behaviour as such I didn't want to make the change in this PR. I think adding a retry on the publisher attempt is worthwhile, but as you say hard failing the entire package list due to one marshal error seems OTT.

Qinusty: Opened https://github.com/ossf/package-feeds/issues/100 and https://github.com/ossf/package-feeds/issues/101 to track these issues.

cmd/scheduled-feed/main.go
tom--pollard: Agreed, changed to `cronPoll()`

cmd/scheduled-feed/main.go
tom--pollard: Yep, thanks for raising https://github.com/ossf/package-feeds/issues/100

cmd/scheduled-feed/main.go
tom--pollard: The bool returned can be blanked however I've tried to match the behaviour `ServerHTTP` here as to not log/write  `processedPackages` if there's any errors.  We could look to always output the value before returning any errors, if it's not zero?

Qinusty (approved):
Looks good to me :) We can address those minor raised issues separately

Issue #93, opened 2021-04-22
Align package types with purl-spec?


georgevanburgh:
Something that was mentioned as part of #26 was the possibility of adding a package-url to the output. This could follow the [`purl-spec`](https://github.com/package-url/purl-spec), which would allow downstream consumers to use other tools that conform to the spec to download new versions of packages for analysis. However, some of the package type names used in this project conflict with the [well-known purl types](https://github.com/package-url/purl-spec#known-purl-types) from the spec. Here's how things compare currently:

| purl-spec  | package-feeds | match? |
| ----------- | ---------------- | -------- |
| cargo | [crates](https://github.com/ossf/package-feeds/blob/4d23df36eb98e7e07713baeb12986bd7f7defdf3/feeds/crates/crates.go#L12) | ❌ |
| composer | [packagist](https://github.com/ossf/package-feeds/blob/4d23df36eb98e7e07713baeb12986bd7f7defdf3/feeds/packagist/packagist.go#L13) | ❌ |
| golang | [goproxy](https://github.com/ossf/package-feeds/blob/4d23df36eb98e7e07713baeb12986bd7f7defdf3/feeds/goproxy/goproxy.go#L15) | ❌ |
| gem | [rubygem](https://github.com/ossf/package-feeds/blob/main/feeds/rubygems/rubygems.go#L13) | ❌ |
| nuget | [nuget](https://github.com/ossf/package-feeds/blob/4d23df36eb98e7e07713baeb12986bd7f7defdf3/feeds/nuget/nuget.go#L13) | ✔ |
| npm | [npm](https://github.com/ossf/package-feeds/blob/4d23df36eb98e7e07713baeb12986bd7f7defdf3/feeds/npm/npm.go#L14) | ✔ |
| pypi | [pypi](https://github.com/ossf/package-feeds/blob/4d23df36eb98e7e07713baeb12986bd7f7defdf3/feeds/pypi/pypi.go#L13) | ✔ |

In general I think this boils down to the `purl-spec` project naming according to the package _type_, whereas we have feeds named after the package _source_. In terms of options, I see a few:

- Update package-feed feeds to match the `purl-spec` types. This would be a one-time breaking change for existing downstream consumers.
[...]

PR #99, opened 2021-04-30
Add support for optional package lists for feeds

Qinusty:
This PR allows feeds to be configured with a set of packages to poll. In the case of certain feeds, this can help avoid [lossy behaviour](https://github.com/ossf/package-feeds/issues/83) by ensuring the latest updates from those specific packages are checked during polling.

Support for this option is currently implemented in pypi and an error is reported when attempting to configure other feeds with this unsupported option.

Partially resolves https://github.com/ossf/package-feeds/issues/83

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -6,16 +6,18 @@ import (
 
 	"github.com/ossf/package-feeds/config"
 	"github.com/ossf/package-feeds/events"
+	"github.com/ossf/package-feeds/feeds"
+	"github.com/ossf/package-feeds/feeds/pypi"
 	"github.com/ossf/package-feeds/feeds/scheduler"
 	"github.com/ossf/package-feeds/publisher/stdout"
 )
 
 const (
 	TestConfigStr = `
-enabled_feeds:
-- rubygems
-- goproxy
-- npm
+feeds:
+- type: rubygems
+- type: goproxy
+- type: npm
 
 publisher:
   type: "gcp"
@@ -27,8 +29,8 @@ poll_rate: 5m
 timer: true
 `
 	TestConfigStrUnknownFeedType = `
-enabled_feeds:
-- foo
+feeds:
+- type: foo
 `
 	TestConfigStrUnknownField = `
 foo:
@@ -52,11 +54,11 @@ func TestDefault(t *testing.T) {
 	t.Parallel()
 
 	c := config.Default()
-	feeds, err := c.GetScheduledFeeds()
+	scheduledFeeds, err := c.GetScheduledFeeds()
 	if err != nil {
 		t.Fatalf("failed to initialize feeds: %v", err)
 	}
-	_ = scheduler.New(feeds)
+	_ = scheduler.New(scheduledFeeds)
 }
 
 func TestGetScheduledFeeds(t *testing.T) {
@@ -66,16 +68,16 @@ func TestGetScheduledFeeds(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(c.EnabledFeeds) != 3 {
-		t.Fatalf("EnabledFeeds is expected to be 3 but was `%v`", len(c.EnabledFeeds))
+	if len(c.Feeds) != 3 {
+		t.Fatalf("Feeds is expected to be 3 but was `%v`", len(c.Feeds))
 	}
-	feeds, err := c.GetScheduledFeeds()
+	scheduledFeeds, err := c.GetScheduledFeeds()
 	if err != nil {
 		t.Fatal(err)
 	}
-	for _, val := range c.EnabledFeeds {
-		if _, ok := feeds[val]; !ok {
-			t.Errorf("expected `%v` feed was not found in scheduled feeds after GetScheduledFeeds()", val)
+	for _, feed := range c.Feeds {
+		if _, ok := scheduledFeeds[feed.Type]; !ok {
+			t.Errorf("expected `%v` feed was not found in scheduled feeds after GetScheduledFeeds()", feed.Type)
 		}
 	}
 }
@@ -93,7 +95,7 @@ func TestLoadFeedConfigUnknownFeedType(t *testing.T) {
 	}
 }
 
-func TestPubConfigToPublisherStdout(t *testing.T) {
+func TestPublisherConfigToPublisherStdout(t *testing.T) {
 	t.Parallel()
 
 	c := config.PublisherConfig{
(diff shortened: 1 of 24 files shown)


config/config_test.go
```diff
@@ -97,6 +99,46 @@ func TestPubConfigToPublisherStdout(t *testing.T) {
...
+
+	c := config.FeedConfig{
+		Type: pypi.FeedName,
+		Options: feeds.FeedOptions{
+			Packages: &packages,
+		},
+	}
+	feed, err := c.ToFeed()
+	if err != nil {
+		t.Fatalf("failed to create pypi critical feed from configuration: %v", err)
```
Qinusty, line 119: Remains of previous `pypi_critical` test case. Will remove.

config/config_test.go
```diff
@@ -97,6 +99,46 @@ func TestPubConfigToPublisherStdout(t *testing.T) {
...
+	}
+
+	pypiFeed, ok := feed.(*pypi.Feed)
+	if !ok {
+		t.Fatal("failed to cast feed as pypi critical feed")
```
Qinusty, line 124: Remains of previous `pypi_critical` test case. Will remove.

config/config_test.go
```diff
@@ -97,6 +99,46 @@ func TestPubConfigToPublisherStdout(t *testing.T) {
...
+	}
+
+	feedPackages := pypiFeed.GetPackageList()
+	if feedPackages == nil {
+		t.Fatalf("failed to initialize pypi feed package list to poll")
+	}
+	if feedPackages != nil && len(*feedPackages) != len(packages) {
+		t.Errorf("pypi critical package list does not match config provided package list")
```
Qinusty, line 132: Remains of previous `pypi_critical` test case. Will remove.

config/config_test.go
```diff
@@ -97,6 +99,46 @@ func TestPubConfigToPublisherStdout(t *testing.T) {
...
+	} else {
+		for i := 0; i < len(packages); i++ {
+			if (*feedPackages)[i] != packages[i] {
+				t.Errorf("pypi critical package '%v' does not match configured package '%v'", (*feedPackages)[i], packages[i])
```
Qinusty, line 136: Remains of previous `pypi_critical` test case. Will remove.

config/structs.go
```diff
@@ -1,14 +1,23 @@
...
-	EnabledFeeds []string        `yaml:"enabled_feeds"`
-	HTTPPort     int             `yaml:"http_port,omitempty"`
-	PollRate     string          `yaml:"poll_rate"`
-	Timer        bool            `yaml:"timer"`
+	PubConfig PublisherConfig `yaml:"publisher"`
+	Feeds     []FeedConfig    `yaml:"feeds"`
+	HTTPPort  int             `yaml:"http_port,omitempty"`
+	PollRate  string          `yaml:"poll_rate"`
+	Timer     bool            `yaml:"timer"`
 }
```
Qinusty, line 13: The change here breaks configuration for anyone who has been using the `enabled_feeds` configuration option since added.

Keeping support for this is possible however it complicates the code unnecessarily as the migration to the new configuration is a minimal change.

```

enabled_feeds:
- pypi
```

Becomes 

```
feeds:
- type: pypi
```

config/scheduledfeed.go
```diff
@@ -135,9 +143,12 @@ func (pc PublisherConfig) ToPublisher(ctx context.Context) (publisher.Publisher,
 	case stdout.PublisherType:
 		return stdout.New(), nil
 	default:
-		err = fmt.Errorf("%w : %v", errUnknownPub, pc.Type)
+		return nil, fmt.Errorf("%w : %v", errUnknownPub, pc.Type)
 	}
-	return nil, err
+}
+
+func (fc FeedConfig) ToFeed() (feeds.ScheduledFeed, error) {
```
tom--pollard, line 150: Is there a need to have the public constructor behind this func?

feeds/crates/crates.go
```diff
@@ -52,6 +52,16 @@ func fetchPackages() ([]*Package, error) {
 
 type Feed struct{}
 
+func New(feedOptions feeds.FeedOptions) (*Feed, error) {
+	if feedOptions.Packages != nil {
```
tom--pollard, line 56: I wonder if we can extract the guts of this into `UnsupportedOption()` or similar func in `feeds`?  Maybe passing in a const/globalvar of what the calling feed type supports. Would mean only one func needs to be updated as potential feed options expand?

Qinusty: Potentially, it might make sense for feeds to specify some map of supported fields with a helper function to check for these. But until more options are added I think the current implementation is sufficient. I'll leave it to whoever expands the Options to decide the best way to tackle this for a multi-option approach.

config/scheduledfeed.go
Qinusty: I think this remains after some shifting around of code, the wrapped function was used elsewhere without a FeedConfig, but this is no longer needed so I'll simplify this down to just `ToFeed()`.

config/structs.go
tom--pollard: Speaking of breaking changes, there's an ongoing issue around if the value for the `Package{}` `type` field should be changed, or even the keyword itself https://github.com/ossf/package-feeds/issues/93

If the latter is the case then this may inadvertently require another breaking change to this option.

feeds/crates/crates.go
tom--pollard: Yes it's definitely something that could be implemented at a later stage, and in a non breaking manner!

feeds/pypi/pypi.go
```diff
@@ -72,20 +86,94 @@ func fetchPackages() ([]*Package, error) {
...
+	}, nil
+}
 
 func (feed Feed) Latest(cutoff time.Time) ([]*feeds.Package, error) {
 	pkgs := []*feeds.Package{}
-	pypiPackages, err := fetchPackages()
+	var pypiPackages []*Package
+	var err error
+
+	if feed.packages == nil {
```
tom--pollard, line 148: For a scenario where somebody is interested in the firehose as-well as set of specific packages, the expectation would be with this implementation that they'd have to run two instances of the binary? This make the configuration logic clean, but do we want to consider allowing either/or here for a single instance?

Qinusty: Currently it is only possible to have one feed of each type. In a previous PR this was possible but increased complexity of the code and configuration. We should see if there is much interest in this before looking to implement this.

tom--pollard: Sure, makes sense to tackle it if requested

feeds/pypi/pypi.go
```diff
@@ -73,28 +87,97 @@ func fetchPackages() ([]*Package, error) {
...
+	errs := make(chan error)
+	var wg sync.WaitGroup
+
+	for _, pkgName := range packageList {
+		wg.Add(1)
+		go func(pkgName string) {
+			resp, err := httpClient.Get(fmt.Sprintf(packageURLFormat, pkgName))
+			if err != nil {
+				errs <- err
+				return
```
tom--pollard, line 101: Does this need an `defer wg.Done()` at the start of the func (or explicitly on each error return) to notify the later `wg.Wait()` to stop blocking on `return`? I'm not sure if this behaviour is triggered by the return statement already.

feeds/pypi/pypi.go
```diff
@@ -73,28 +87,97 @@ func fetchPackages() ([]*Package, error) {
 	return rssResponse.Packages, nil
 }
 
+func fetchCriticalPackages(packageList []string) ([]*Package, error) {
+	responseChannel := make(chan *Response)
```
tom--pollard, line 91: Both of these channels are non buffered, afaict this will lead to the later `select` statements only processing the last value written to them?

feeds/pypi/pypi.go
```diff
@@ -73,28 +87,97 @@ func fetchPackages() ([]*Package, error) {
...
+			return nil, err
+		}
+		pkgVersion, err := pkg.Version()
+		if err != nil {
+			return nil, err
+		}
+		pkg := feeds.NewPackage(pkg.CreatedDate.Time, pkgName, pkgVersion, FeedName)
 		pkgs = append(pkgs, pkg)
 	}
 	feed.lossyFeedAlerter.ProcessPackages(FeedName, pkgs)
```
tom--pollard, line 175: Checking for overlaps could be a little redundant when configured for `CriticalPackages` as every configured package would have had to have 50 (I assume the package release RSS is also limited to 50) releases in-between a polling event to not cause an overlap. It may make more sense to check old-new results on a per package basis with this in mind, or not at all.

feeds/pypi/pypi.go
```diff
@@ -73,28 +85,93 @@ func fetchPackages() ([]*Package, error) {
...
+	}
+
+	pkgs := []*Package{}
+	var lastErr error
+	for i := 0; i < len(packageList); i++ {
+		select {
+		case response := <-responseChannel:
+			pkgs = append(pkgs, response.Packages...)
+		case err := <-errChannel:
+			lastErr = err
```
tom--pollard, line 118: As `Latest()` returns no packages if there's an error given from this func, it'd be quicker just to `return nil, err` immediately here if an error is detected. Unless there's inherent value in the last error, as opposed to the first.

tom--pollard (approved):

Issue #102, opened 2021-05-05
Improve npm polling performance


Qinusty:
Currently the npm feed queries an initial endpoint to get a list of packages which have been updated, and then follows up with queries to the individual packages to retrieve the version information of the update. This is done sequentially and therefore is a significant bottleneck for any deployment polling npm.

PR #104, opened 2021-05-06
Utilise goroutines when polling for npm package versions

tom--pollard:
Resolves https://github.com/ossf/package-feeds/issues/102

With `npm` as the only configured feed, some rudimentary results when executed side-by-side with **main** on the same host.

**main-bc1e1e9**
```
time curl http://localhost:8080
49 packages processed
real	0m6.562s
user	0m0.010s
sys	0m0.005s

```
**tom--pollard:npm_goroutine**
```
time curl http://localhost:8081
49 packages processed
real	0m1.410s
user	0m0.011s
sys	0m0.000s
```

### DIFF
--- a/feeds/npm/npm.go
+++ b/feeds/npm/npm.go
@@ -5,6 +5,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
+	"sort"
 	"time"
 
 	"github.com/ossf/package-feeds/events"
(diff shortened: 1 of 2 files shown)


feeds/npm/npm.go
```diff
@@ -102,18 +102,35 @@ func New(eventHandler *events.Handler) *Feed {
...
+		case pkg := <-packageChannel:
+			pkgs = append(pkgs, pkg)
+		case err := <-errs:
 			return pkgs, fmt.Errorf("error in fetching version information: %w", err)
 		}
-		pkg := feeds.NewPackage(pkg.CreatedDate.Time, pkg.Title, v, FeedName)
-		pkgs = append(pkgs, pkg)
 	}
+
 	feed.lossyFeedAlerter.ProcessPackages(FeedName, pkgs)
```
Qinusty, line 134: How does the NPM API work for the version information? Is there chance of overlap detection for lossy feeds?

tom--pollard: The package specific endpoint only provides the `latest` version, so without known previous version state or enforcement of versioning patterns then I think it's worth doing the overlap check.

tom--pollard: Yep, looks like there's a fundamental problem with the package logic here with npm parsing. Will be tackled via #106

Qinusty (approved):
LGTM!

PR #113, opened 2021-05-13
Ignore non-utf8 in XML decoding

### LINKED ISSUE #112: Invalid UTF-8 characters cause failures in polling RSS feeds
This is due to the XML unmarshal function being unable to parse these symbols.

We should ensure the rss response body contains only supported characters and any invalid characters should be stripped/replaced.

Qinusty:
Resolves #112 

Strips UTF-8 replacement characters from readers provided to XML decoders throughout feeds.

### DIFF
--- a/feeds/npm/npm.go
+++ b/feeds/npm/npm.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ossf/package-feeds/events"
 	"github.com/ossf/package-feeds/feeds"
+	"github.com/ossf/package-feeds/utils"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


feeds/npm/npm.go
```diff
@@ -95,6 +91,8 @@ func fetchVersionInformation(packageName string) (string, error) {
 
 type Feed struct {
 	lossyFeedAlerter *feeds.LossyFeedAlerter
+	firehoseURL      string
```
tom--pollard, line 94: Might want to keep this called `baseURL`, seems more flexible and easier to understand if someone is unaware of usecase here of `firehose`

Qinusty: I'm keen to rename these during these changes, perhaps not to firehose but the terminology of baseURL just seems incorrect as it's usually a URL containing the full path. I'll play around with the naming a little.

Qinusty: baseURL is now actually the baseURL and the paths are added on where relevant. This makes hooking the tests up slightly more straight forward too since you can just provide the test server URL.

tom--pollard:
Shouldn't the commit title be 'Ignore **non**-utf8 in XML decoding' ? Other than that LGTM

tom--pollard (approved):

PR #115, opened 2021-05-13
Utilise Feed structs for URL definitions

Qinusty:
Previously multiple package tests couldn't be ran in
parallel due to the package level variables for URLs.
With these changes many package tests can be ran in
parallel without interfering with eachother.


tom--pollard (approved):
LGTM, I see no issue with keeping it on the struct instances instead of globally across the package.

PR #119, opened 2021-05-17
Add StatusCode checks to feed HTTP requests

### LINKED ISSUE #114: Check status codes prior to decoding response body
All feeds currently attempt to decode a response body which can lead to somewhat unclear error messages, decoding should not occur for status codes outside of 200-299 and appropriate errors should be returned and logged.

Qinusty:
This resolves #114, improving the clarity on error messages produced when a feed hits a non successful status in a response.


tom--pollard (approved):
LGTM! Maybe some future work could go towards retries on err responses

PR #121, opened 2021-05-19
WIP: Refactor scheduler/polling out of scheduled-feed/main.go

Qinusty:
scheduled-feed/main.go was becoming quite cluttered as functionality has been added, this PR remedies this by shifting many of the responsibilities to the scheduler with additional structs being introduced to handle the polling of feed collections.


feeds/scheduler/feed_group.go
```diff
@@ -0,0 +1,115 @@
...
+		log.Errorf("Failed to publish %v packages due to err: %v", len(pkgs)-numPublished, err)
+		return numPublished, err
+	}
+	log.WithField("num_packages", numPublished).Printf("Successfully published packages")
+	return numPublished, nil
+}
+
+// Poll fetches the latest packages from each registered feed.
+func (fg *FeedGroup) poll() ([]*feeds.Package, []error) {
+	results := make(chan pollResult)
```
tom--pollard, line 56: I think we may want to buffer the channel `make(chan pollResult, len(fg.feeds))`

feeds/scheduler/feed_group.go
```diff
@@ -0,0 +1,115 @@
...
+	publisher publisher.Publisher
+	lastPoll  time.Time
+}
+
+func NewFeedGroup(scheduledFeeds []feeds.ScheduledFeed,
+	pub publisher.Publisher, initialCutoff time.Duration) *FeedGroup {
+	return &FeedGroup{
+		feeds:     scheduledFeeds,
+		publisher: pub,
+		lastPoll:  time.Now().Add(-initialCutoff),
```
tom--pollard, line 25: Should probably be explicit with `.UTC()`

feeds/scheduler/poll_handler.go
```diff
@@ -0,0 +1,48 @@
...
+	err          error
+}
+
+func NewPollHandler(feeds []*FeedGroup) *PollHandler {
+	return &PollHandler{feedGroups: feeds}
+}
+
+func (srv *PollHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var err error
+	resultChannel := make(chan publishResult)
```
tom--pollard, line 23: Same comment before re buffering

feeds/scheduler/poll_handler.go
```diff
@@ -0,0 +1,48 @@
+package scheduler
+
+import (
+	"fmt"
+	"net/http"
+)
+
+type PollHandler struct {
```
tom--pollard, line 8: I'm not entirely sure why but having this file/struct named Poll Handler feels possibility a little confusing, on first skim I assumed it would contain the logic for polling the feeds, not accepting the external polls to the package's http server. The name does make sense, I just wonder if we could avoid the confusion?

Qinusty: The justification was Handler for [http.Handler](https://golang.org/pkg/net/http/#Handler) and Poll relating to in fact polling the feeds, as FeedGroup is used to perform the actual polling operations. Open to suggestions though.

tom--pollard:
Other than my previous comments, this looks a lot cleaner! It also opens up a smoother path to introducing per feed polling rates, as mentioned in #85

feeds/scheduler/poll_handler.go
tom--pollard: Yeh I understand, if there wasn't other polling operations within the system then there'd be no potential confusion at all. I'm struggling to think of a suitable suggestion, but maybe something appended/prepended. Obviously this is not a blocker.

tom--pollard:
Running the binary locally, and triggering via a remote curl, at the end of the logs I get:

```
time="2021-05-20T11:17:11Z" level=info msg="Packages successfully processed" feed=pypi num_processed=0
time="2021-05-20T11:17:11Z" level=info msg="Packages successfully processed" feed=goproxy num_processed=0
time="2021-05-20T11:17:11Z" level=info msg="Packages successfully processed" feed=packagist num_processed=0
time="2021-05-20T11:17:12Z" level=info msg="Packages successfully processed" feed=crates num_processed=0
time="2021-05-20T11:17:13Z" level=info msg="Packages successfully processed" feed=nuget num_processed=0
time="2021-05-20T11:17:13Z" level=info msg="Packages successfully processed" feed=npm num_processed=0
time="2021-05-20T11:17:13Z" level=info msg="Packages successfully processed" feed=rubygems num_processed=0
time="2021-05-20T11:17:13Z" level=info msg="0 packages processed"
```
I've noticed before these changes that the `num_processed` logging can sometimes be odd, but we should probably look to address this in the PR.

feeds/scheduler/scheduler.go
```diff
@@ -27,39 +35,35 @@ type pollResult struct {
...
-			log.WithFields(log.Fields{
-				"feed":    result.name,
-				"name":    pkg.Name,
-				"version": pkg.Version,
-			}).Print("Processing Package")
+	// Configure cron job for scheduled polling
+	cronJob := cron.New()
+	feedGroup := NewFeedGroup(allFeeds, s.publisher, initialCutoff)
+	if enableDefaultTimer {
+		err := cronJob.AddJob(defaultSchedule, feedGroup)
```
tom--pollard, line 53: Main currently logs/prints that it's running a timer at a given schedule, can we add this here too?

Qinusty:
> I've noticed before these changes that the `num_processed` logging can sometimes be odd, but we should probably look to address this in the PR.


num_processed=0 is due to no packages being processed within the cutoff window I believe. I'm not witnessing this with longer pollRate values.

tom--pollard:
> > I've noticed before these changes that the `num_processed` logging can sometimes be odd, but we should probably look to address this in the PR.
> 
> num_processed=0 is due to no packages being processed within the cutoff window I believe. I'm not witnessing this with longer pollRate values.

I'm not sure that's the case,  on a clean execution with the default time, I get multiple `sending package upstream` logs, and a return of `n packages processed` on the http request.

feeds/scheduler/poll_handler.go
Qinusty: I've gone with `FeedGroupsHandler` to be more verbose, this lines up with serving data relating to the operations of the `FeedGroups`.

tom--pollard (approved):
LGTM!

PR #122, opened 2021-05-20
Add support for individual feed poll timers

Qinusty:
This follows on from https://github.com/ossf/package-feeds/issues/85 to introduce separate polling timers for individual feeds.

Feeds are grouped and polled based on their schedules defined in FeedOptions, alternatively a default schedule is applied based upon pollRate in the root of the configuration (which runs if `timer` is enabled in the configuration).

### DIFF
--- a/feeds/README.md
+++ b/feeds/README.md
@@ -1,0 +1,44 @@
+# Feeds
+
+Each of the feeds have their own implementation and support their own set of configuration options.
+
+## Configuration options
+
+`packages` this configuration option is only available on certain feeds, check the README of the feed you're interested in for information on this.
+
+`poll_rate` this allows for setting the frequency of polling for this specific feed. This is supported by all feeds. The value should be a string formatted for [duration parser](https://golang.org/pkg/time/#ParseDuration). Setting this value will enable the scheduled polling regardless of the value of `timer` in the root of the configuration.
+
+## Example
+
+### Poll Pypi every 5 minutes
+
+```
+feeds:
+- type: pypi
+  options:
+    poll_rate: "5m"
+```
+
+### Poll npm every 10 minutes and crates every hour
+
+```
+feeds:
+- type: npm
+  options:
+    poll_rate: "5m"
+- type: crates
+  options:
+    poll_rate: "1h"
+```
+
+### Poll a subset of pypi every 10 minutes
+
+```
+feeds:
+- type: pypi
+  options:
+    packages:
+    - numpy
+    - django
+    poll_rate: "10m"
+```
--- a/feeds/scheduler/scheduler.go
+++ b/feeds/scheduler/scheduler.go
@@ -3,6 +3,7 @@ package scheduler
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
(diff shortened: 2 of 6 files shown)


feeds/README.md
```diff
@@ -0,0 +1,44 @@
+# Feeds
+
+Each of the feeds have their own implementation and support their own set of configuration options.
+
+## Configuration options
+
+`packages` this configuration option is only available on certain feeds, check the README of the feed you're interested in for information on this.
+
+`schedule` this allows for polling to be configured on a schedule for this specific feed. This is supported by all feeds. The value for this field should match the format defined by the [cron library](https://pkg.go.dev/github.com/robfig/cron#hdr-CRON_Expression_Format), intervals and predefined schedules are also supported.
```
tom--pollard, line 9: Having the main option of  `poll_rate` parsed as time.Duration and feed specific schedules as cron formatted strings might be slightly confusing for users. Do we want to be consistent here? `poll_rate` has the caveat that it's used as an actual duration value too for the cutoff.

Qinusty: Hmmm... perhaps it's worth restricting this value to be intervals set as time.Duration parsable strings which we can prepend with `@every`.

I think this is flexible enough for most usecases.

tom--pollard: I'd vote for the prepending where necessary :)

feeds/scheduler/scheduler.go
```diff
@@ -68,3 +81,21 @@ func (s *Scheduler) Run(initialCutoff time.Duration, enableDefaultTimer bool) er
...
+func buildSchedules(registry map[string]feeds.ScheduledFeed, pub publisher.Publisher,
+	initialCutoff time.Duration) map[string]*FeedGroup {
+	schedules := map[string]*FeedGroup{}
+	for _, feed := range registry {
+		options := feed.GetFeedOptions()
+		schedule := options.Schedule
+
+		// Initialize new schedules in map.
+		if _, ok := schedules[schedule]; !ok {
+			schedules[schedule] = NewFeedGroup([]feeds.ScheduledFeed{}, pub, initialCutoff)
```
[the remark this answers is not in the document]
Qinusty: cutoff is now set based upon the poll_rate option provided to feed config (previously named schedule)

feeds/README.md
Qinusty: renamed `schedule` to `poll_rate` and changed logic to operate identically to poll_rate in the top level configuration.

feeds/scheduler/scheduler.go
```diff
@@ -41,25 +42,37 @@ type pollResult struct {
...
+
+		// Undefined schedules will follow the default schedule, if the default timer is enabled.
+		if schedule == "" {
+			if !enableDefaultTimer {
+				continue
+			}
+			schedule = defaultSchedule
+		}
+
+		err := cronJob.AddJob(schedule, feedGroup)
```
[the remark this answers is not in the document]
Qinusty: Updated docs to clarify this behaviour, naming remains the same.

tom--pollard (approved):

Issue #101, opened 2021-05-04
Failing to marshal a package should not result in hard failure during publishing


Qinusty:
Failing to marshal a single package currently blocks the publishing of all subsequent packages, this is undesirable. 

https://github.com/ossf/package-feeds/blob/ce53f3ea/cmd/scheduled-feed/main.go#L45-L50 is the relevant code chunk.

tom--pollard:
Seems sensible that this shouldn't lead to a complete loss of the subsequent package data. This might be somewhere to add another event, and simply continue.

tom--pollard:
Now that #123 has merged I'll look to update `publishPackages()` to handle errors on a per package basis, using `errPub` to align with `poll()`

PR #123, opened 2021-05-21
 feeds/: Allow for per package/feed error handling when polling

### LINKED ISSUE #107: Handle errors more granularly when polling for specific packages
With https://github.com/ossf/package-feeds/pull/99 landing, the ability to provide a specific set of packages for polling was added to the `pypi` feed, with the framework for this to be extended to other similar feeds.

This brings with it the caveat that if when polling a feed for any of the set of packages fails (e.g, network error/timeout), then the whole set of package data from that specific polling session is lost. This is due to the assumption from the generic polling of the 'firehose' that only one roundtrip will be performed, as oppose to the N that may be done if watching a given set of endpoints. 

- https://github.com/ossf/package-feeds/blob/main/feeds/pypi/pypi.go#L149 

- https://github.com/ossf/package-feeds/blob/main/feeds/scheduler/scheduler.go#L49

While it's right to report the error/errors, if a partial set of packages were successfully polled then it's reasonable to continue processing them and could be a candidate for the `event` framework.

tom--pollard:
Resolves #107 

    feeds/: Allow for per package/feed error handling when polling
    
    When errors occur but package data can still be processed then
    the system should continue where possible. This is particularly
    beneficial in feeds where data is fetched on a per package basis,
    as a singular failure shouldn't immediately lead to data loss
    for the whole session. `pollAndPublish()` error handling and
    subsequent callers have been modified to handle this.
    
    This also removes `errUnpublished` as a 'hard' error when polling
    npm for critical packages as such it is only logged.

Follow ups in the form of #101 and more tests/granular handling in nuget & packagist should also be undertaken

### DIFF
--- a/feeds/feed.go
+++ b/feeds/feed.go
@@ -1,11 +1,14 @@
 package feeds
 
 import (
+	"errors"
 	"fmt"
 	"time"
 )
 
 const schemaVer = "1.0"
+
+var ErrNoPackagesPolled = errors.New("no packages were successfully polled")
 
 type UnsupportedOptionError struct {
 	Option string
@@ -13,7 +16,7 @@ type UnsupportedOptionError struct {
 }
 
 type ScheduledFeed interface {
-	Latest(cutoff time.Time) ([]*Package, error)
+	Latest(cutoff time.Time) ([]*Package, []error)
 	GetFeedOptions() FeedOptions
 	GetName() string
 }
@@ -35,6 +38,15 @@ type Package struct {
 	CreatedDate time.Time `json:"created_date"`
 	Type        string    `json:"type"`
 	SchemaVer   string    `json:"schema_ver"`
+}
+
+type PackagePollError struct {
+	Err  error
+	Name string
+}
+
+func (err PackagePollError) Error() string {
+	return fmt.Sprintf("Polling for package %s returned error: %v", err.Name, err.Err)
 }
 
 func NewPackage(created time.Time, name, version, feed string) *Package {
--- a/feeds/pypi/pypi.go
+++ b/feeds/pypi/pypi.go
@@ -97,7 +97,7 @@ func fetchPackages(baseURL string) ([]*Package, error) {
 	return rssResponse.Packages, nil
 }
 
-func fetchCriticalPackages(baseURL string, packageList []string) ([]*Package, error) {
+func fetchCriticalPackages(baseURL string, packageList []string) ([]*Package, []error) {
 	responseChannel := make(chan *Response)
 	errChannel := make(chan error)
 
@@ -106,19 +106,19 @@ func fetchCriticalPackages(baseURL string, packageList []string) ([]*Package, er
 			packageDataPath := fmt.Sprintf(packagePathFormat, pkgName)
 			pkgURL, err := utils.URLPathJoin(baseURL, packageDataPath)
 			if err != nil {
-				errChannel <- err
+				errChannel <- feeds.PackagePollError{Name: pkgName, Err: err}
 				return
 			}
 			resp, err := httpClient.Get(pkgURL)
 			if err != nil {
-				errChannel <- err
+				errChannel <- feeds.PackagePollError{Name: pkgName, Err: err}
 				return
 			}
 			defer resp.Body.Close()
 
 			err = utils.CheckResponseStatus(resp)
 			if err != nil {
-				errChannel <- fmt.Errorf("failed to fetch pypi package data: %w", err)
+				errChannel <- feeds.PackagePollError{Name: pkgName, Err: fmt.Errorf("failed to fetch pypi package data: %w", err)}
 				return
 			}
 
@@ -126,7 +126,7 @@ func fetchCriticalPackages(baseURL string, packageList []string) ([]*Package, er
 			reader := utils.NewUTF8OnlyReader(resp.Body)
 			err = xml.NewDecoder(reader).Decode(rssResponse)
 			if err != nil {
-				errChannel <- err
+				errChannel <- feeds.PackagePollError{Name: pkgName, Err: err}
 				return
 			}
 
@@ -135,15 +135,16 @@ func fetchCriticalPackages(baseURL string, packageList []string) ([]*Package, er
 	}
 
 	pkgs := []*Package{}
+	errs := []error{}
 	for i := 0; i < len(packageList); i++ {
 		select {
 		case response := <-responseChannel:
 			pkgs = append(pkgs, response.Packages...)
 		case err := <-errChannel:
-			return nil, err
+			errs = append(errs, err)
 		}
 	}
-	return pkgs, nil
+	return pkgs, errs
 }
 
 type Feed struct {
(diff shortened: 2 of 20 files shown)


tom--pollard:
If it's deemed necessary I can look to add this behaviour for all feeds in a single PR

feeds/pypi/pypi.go
```diff
@@ -167,19 +170,30 @@ func New(feedOptions feeds.FeedOptions, eventHandler *events.Handler) (*Feed, er
...
 	} else {
 		// Fetch specific packages individually from configured packages list.
-		pypiPackages, err = fetchCriticalPackages(feed.baseURL, *feed.packages)
+		pypiPackages, failedPackages = fetchCriticalPackages(feed.baseURL, *feed.packages)
+		for _, pkg := range failedPackages {
+			// This could be an entry point for a 'lossy' type of event.
+			log.Errorf("Failed to poll for %s package due to err: %v", pkg.Name, pkg.Err)
+		}
+		if len(pypiPackages) == 0 {
+			// If none of the packages were successfully polled for, return early.
```
Qinusty, line 192: I'm not sure returning early is necessary as the following code blocks don't do anything if there are no pypi packages. Additionally that means we don't need the `errors.Is` check.

tom--pollard: The `errors.Is` check was more for testing purposes to ensure that the correct error was propagated to the user. If we don't think that is necessary then I agree the early return isn't needed.

feeds/feed.go
```diff
@@ -37,6 +40,11 @@ type Package struct {
 	SchemaVer   string    `json:"schema_ver"`
 }
 
+type FailedPackagePoll struct {
```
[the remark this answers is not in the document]
tom--pollard: Yeh I'll take a look at doing it this way, and for further feeds where it's appropriate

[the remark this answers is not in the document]
Qinusty: Yeah I think it's definitely worth adding to the other feeds as we're currently risking the loss of a bunch of packages we successfully polled!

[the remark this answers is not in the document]
tom--pollard: I've provided the `Error()` interface on this struct, hopefully inline with what you were suggesting :)

feeds/pypi/pypi.go
```diff
@@ -168,18 +171,29 @@ func (feed Feed) Latest(cutoff time.Time) ([]*feeds.Package, error) {
...
 		pypiPackages, err = fetchPackages(feed.baseURL)
+		if err != nil {
+			return nil, err
+		}
 	} else {
 		// Fetch specific packages individually from configured packages list.
-		pypiPackages, err = fetchCriticalPackages(feed.baseURL, *feed.packages)
+		pypiPackages, errs = fetchCriticalPackages(feed.baseURL, *feed.packages)
+		for _, err := range errs {
+			// This could be an entry point for a 'lossy' type of event.
```
Qinusty, line 188: Should this be a TODO?

feeds/feed.go
```diff
@@ -37,6 +40,15 @@ type Package struct {
...
```
Qinusty, line 43: We should name this more conventionally including `Error` in the name

feeds/pypi/pypi.go
tom--pollard: Yep!

feeds/feed.go
tom--pollard: Agreed, changed to  `PackagePollError`

tom--pollard:
@Qinusty In terms of beyond general review, I'm particularly interested in ensuring nuget/packagist are ok as it's my first time working with those feed implementations. I'm also happy to split any **new** tests into a separate commit once reviewed.

feeds/scheduler/feed_group.go
```diff
@@ -38,10 +38,9 @@ func (fg *FeedGroup) Run() {
 }
 
 func (fg *FeedGroup) PollAndPublish() (int, error) {
-	pkgs, errs := fg.poll()
-	if len(errs) > 0 {
-		return 0, errs[0]
-	} else if len(pkgs) == 0 {
+	// poll() logs errs, so no need to handle.
```
Qinusty, line 41: if `poll()` handles errors then they shouldn't be passed down. Not directly introduced by this PR but we could fix this whilst reworking :).

Qinusty (commented):
Mostly LGTM, if we can fix this outstanding issue it looks good to go

feeds/scheduler/feed_group.go
Qinusty: I think that removing the tests (or this check) is fine if we're removing this pass down of errors. Tests should verify the original feeds output the errors which is handled on a feed by feed basis. We could introduce a parameterized `Logger` throughout which would allow us to provide a [NewNullLogger()](https://github.com/sirupsen/logrus/blob/master/hooks/test/test.go#L42) as the `Logger` for use within the `FeedGroup`, but I think this is unnecessarily complex for the current test scenario.

feeds/scheduler/feed_group.go
```diff
@@ -11,12 +12,23 @@ import (
 	"github.com/ossf/package-feeds/publisher"
 )
 
+var (
+	errPoll = errors.New("error when polling for packages, see log for details")
```
Qinusty, line 16: I'd prefer we avoided mentioning to check the logs in the error itself as this could end up being processed as a go error. "see log for details" should be specific to the HTTP error messaging.

feeds/scheduler/feed_group.go
```diff
@@ -31,28 +43,36 @@ func (fg *FeedGroup) AddFeed(feed feeds.ScheduledFeed) {
...
+func (fg *FeedGroup) pollAndPublish() groupResult {
+	result := groupResult{}
 	pkgs, errs := fg.poll()
-	if len(errs) > 0 {
-		return 0, errs[0]
-	} else if len(pkgs) == 0 {
-		return 0, nil
-	}
 
+	if len(errs) != 0 {
```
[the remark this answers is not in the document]
Qinusty: Accidentally deleted the previous comment... We should remove the logging [here](https://github.com/ossf/package-feeds/pull/123/files#diff-95d0a7cd43229270ff6fe6008662c35352c7819fe8be4f7e313e5b5357281ba0R97) and move it to a higher level to avoid returning errors which have already been logged.

The error list can be handled without the loop via `errs = append(errs, result.errs...)`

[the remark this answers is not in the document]
tom--pollard: I've updated this such that `poll()` still logs the errors, but now it only returns `errPoll/nil`. Keeping the errors logged within poll allows them to be easily logged with the feedname fields, without further wrapping. As a followup, when tackling #101 `publishPackages` should also be modified to log the errors and only return `errPub/nil'

feeds/scheduler/feed_group.go
tom--pollard: Agreed, I've removed "see log for details" from the generic error and added it explicitly to the http error string

Qinusty (commented):
LGTM!

Qinusty (approved):

Issue #125, opened 2021-05-21
Sporadic test failure: feed_test.go:TestValidSchema


Qinusty:
Test occasionally fails with

```
feed_test.go:43: The Package json is not valid against the current schema. see errors :
        - schema_ver: Does not match pattern '^[1-9][0-9]*\.[0-9]+'
```

PR #127, opened 2021-05-21
feeds/feed_test.go: Ensure package structs are thread-safe

tom--pollard:
Resolves https://github.com/ossf/package-feeds/issues/125


Qinusty (approved):
LGTM!

PR #132, opened 2021-06-11
feeds/scheduler/feed_group.go: Handle per package errs when publishing

tom--pollard:
publishPackages() no longer immediately returns if a single package
fails to be marshalled or sent to the publisher. Instead the error
is logged and a generic `errPub` is returned, inline with poll().

This is for #101, however it could be extended in the future with retry logic, or even some form of `event`.

### DIFF
--- a/pkg/scheduler/feed_group.go
+++ b/pkg/scheduler/feed_group.go
@@ -61,13 +61,9 @@ func (fg *FeedGroup) pollAndPublish() groupResult {
 		return result
 	}
 	log.WithField("num_packages", len(pkgs)).Printf("Publishing packages...")
-	numPublished, pubErr := fg.publishPackages(pkgs)
-	result.numPublished = numPublished
-	if pubErr != nil {
-		log.Errorf("Failed to publish %v packages due to err: %v", len(pkgs)-numPublished, pubErr)
-		result.pubErr = errPub
-	} else {
-		log.WithField("num_packages", numPublished).Printf("Successfully published packages")
+	result.numPublished, result.pubErr = fg.publishPackages(pkgs)
+	if result.numPublished > 0 {
+		log.WithField("num_packages", result.numPublished).Printf("Successfully published packages")
 	}
 	return result
 }
@@ -117,6 +113,7 @@ func (fg *FeedGroup) poll() ([]*feeds.Package, error) {
 
 func (fg *FeedGroup) publishPackages(pkgs []*feeds.Package) (int, error) {
 	processed := 0
+	errs := []error{}
 	for _, pkg := range pkgs {
 		log.WithFields(log.Fields{
 			"name":         pkg.Name,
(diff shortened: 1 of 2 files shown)


feeds/scheduler/feed_group.go
```diff
@@ -125,14 +124,21 @@ func (fg *FeedGroup) publishPackages(pkgs []*feeds.Package) (int, error) {
...
 		b, err := json.Marshal(pkg)
 		if err != nil {
-			log.Printf("Error marshaling package: %#v", pkg)
-			return processed, err
+			log.WithField("name", pkg.Name).WithError(err).Error("Error marshaling package")
+			errs = append(errs, err)
 		}
 		if err := (fg.publisher).Send(context.Background(), b); err != nil {
-			log.Printf("Error sending package to upstream publisher %v", err)
-			return processed, err
```
tom--pollard, line 133: This could lead to a situation in which every publish call has to wait for the default timeout (unsure what it is) if say the publisher endpoint is down. I'm not sure this needs to be handled more gracefully in this PR, but it's definitely something to consider.

Qinusty: The timeout appears to be publisher specific with no generalised options for configuring this without using the publisher specific constructors. In the case of kafka this uses sarama for the kafka client implementation, this uses 30s defaults for all read/write timeouts. However this may end up being a 30s timeout for EACH package published.

As for risk reduction, handling publishing in requests would lead to different behaviour where the HTTP response is before the data is published.

feeds/scheduler/feed_group.go
```diff
@@ -61,13 +61,11 @@ func (fg *FeedGroup) pollAndPublish() groupResult {
...
 	}
 	log.WithField("num_packages", len(pkgs)).Printf("Publishing packages...")
-	numPublished, pubErr := fg.publishPackages(pkgs)
-	result.numPublished = numPublished
-	if pubErr != nil {
-		log.Errorf("Failed to publish %v packages due to err: %v", len(pkgs)-numPublished, pubErr)
-		result.pubErr = errPub
-	} else {
-		log.WithField("num_packages", numPublished).Printf("Successfully published packages")
+	numPublished, err := fg.publishPackages(pkgs)
```
Qinusty, line 64:
```suggestion
	result.numPublished, result.pubErr := fg.publishPackages(pkgs)
```
NIT: Does this work?

feeds/scheduler/feed_group.go
```diff
@@ -125,14 +124,21 @@ func (fg *FeedGroup) publishPackages(pkgs []*feeds.Package) (int, error) {
...
 		if err != nil {
-			log.Printf("Error marshaling package: %#v", pkg)
-			return processed, err
+			log.WithField("name", pkg.Name).WithError(err).Error("Error marshaling package")
+			errs = append(errs, err)
 		}
 		if err := (fg.publisher).Send(context.Background(), b); err != nil {
-			log.Printf("Error sending package to upstream publisher %v", err)
-			return processed, err
+			log.WithField("name", pkg.Name).WithError(err).Error("Error sending package to upstream publisher")
```
Qinusty, line 131: The only side effect of this is that it could flood the logs with repeated identical error messages 
```
ERRO[0027] Error sending package to upstream publisher   error="pubsub (code=Unknown): kafka: Failed to deliver 1 messages." name=@skylla/rep
INFO[0027] Sending package upstream                      created_date="2021-06-15 12:49:18.2 +0000 UTC" feed=npm name=@docusaurus/utils-validation
ERRO[0027] Error sending package to upstream publisher   error="pubsub (code=Unknown): kafka: Failed to deliver 1 messages." name=@docusaurus/utils-validation
INFO[0027] Sending package upstream                      created_date="2021-06-15 12:49:12.177 +0000 UTC" feed=npm name=@shack-js/template-repo
ERRO[0027] Error sending package to upstream publisher   error="pubsub (code=Unknown): kafka: Failed to deliver 1 messages." name=@shack-js/template-repo
```
Not a blocking comment, just a thought

Qinusty (approved):
Some thoughts but nothing major, looks good to me!

feeds/scheduler/feed_group.go
tom--pollard: You can't do a shorthand `:=` declaration to a predefined struct's field (where as you can to a generic var, see https://gist.github.com/chmike/ac0113afefbc04e67323b4a3688d6b54#-cant-be-used-with-member-fields). However in this case a normal `=` declaration is fine as we're casting both values to the predefined struct so thanks!

Qinusty (approved):
LGTM!

feeds/scheduler/feed_group.go
tom--pollard: Yeh fair point, the names of the packages which failed to be sent could definitely be stored and then logged in a single entry, or grouped by error type. For now I think this is fine though.

PR #135, opened 2021-06-14
WIP: Add initial integration test support

Qinusty:


integration/docker-compose-integration.yml
```diff
@@ -0,0 +1,37 @@
+version: "3"
+services:
+  zookeeper:
+    image: confluentinc/cp-zookeeper
```
tom--pollard, line 4: Is it worth pinning these containers to a known working tag/ref?

integration/test_consumer/check_kafka_output.py
```diff
@@ -0,0 +1,61 @@
...
+    last_poll_success = datetime.now()
+    while True:
+        msg = c.poll(2.0)
+
+        if msg is None:
+            delta = datetime.now() - last_poll_success
+            # Timeout to avoid hanging on poll() loop
+            if delta.total_seconds() > 10:
+                break
+            else: continue
```
tom--pollard, line 48: NIT: `else` isn't explicitly needed

tom--pollard (approved):
LGTM other than a few small comments, as you've pointed out as this 'test' relies on external services then merges shouldn't be gated behind this explicitly passing. However it's a great starting point for finding runtime errors in various parts of the code not covered by unit tests.

Issue #134, opened 2021-06-14
Panic occurring during NPM package polling


Qinusty:
```
feeds_1      | time="2021-06-14T14:07:43Z" level=info msg="Packages successfully processed" feed=goproxy num_processed=28
feeds_1      | time="2021-06-14T14:07:43Z" level=info msg="Packages successfully processed" feed=crates num_processed=0
feeds_1      | panic: runtime error: slice bounds out of range [:2] with capacity 1
feeds_1      | 
feeds_1      | goroutine 2601 [running]:
feeds_1      | github.com/ossf/package-feeds/feeds/npm.fetchAllPackages.func1(0xcf21ed, 0x1b, 0xc0000ac060, 0xc0000ac000, 0xc0014b1f30, 0xe, 0x2)
feeds_1      |  /app/feeds/npm/npm.go:169 +0x225
feeds_1      | created by github.com/ossf/package-feeds/feeds/npm.fetchAllPackages
feeds_1      |  /app/feeds/npm/npm.go:159 +0x296
```

I haven't been able to reproduce this error but it seems to be due to `count` being less than `len(pkgs)` in some rare cases. [/app/feeds/npm/npm.go:169](https://github.com/ossf/package-feeds/blob/afba893cfa2d6796e931188a49b7989ff5537368/feeds/npm/npm.go#L169)

tom--pollard:
I cannot confirm it (npm package registry history is not 'immutable'), however I *think* this may have occurred where 2 package versions are published, and one of them is then deleted (note, this doesn't mean the package is `unpublished`) by the time the package endpoint is scraped. There is no way to determine this from the json payload, a deleted version is just no longer available.

Realistically here we can only safeguard, and maybe report an `event` in the future.

PR #136, opened 2021-06-14
feeds/npm/npm.go: Guard against 'silent' version removal

tom--pollard:
This should cover the scenario where the number of package specific
'firehose' events is greater than the number of versions available
at process time. Applying the `cutoff` should prevent 'old' versions
being processed in place of the versions that triggered the events.

Resolves https://github.com/ossf/package-feeds/issues/134


Qinusty (approved):
LGTM!

PR #137, opened 2021-06-15
Restructure repository packages into pkg subdirs

### LINKED ISSUE #128: Restructure sub packages into pkg/ folder
The top level of the repository is slightly cluttered with packages, we should tidy these to follow a [standardised directory system](https://github.com/golang-standards/project-layout#go-directories)

Qinusty:
This restructures the repository to reduce the clutter in the root directory of the repository. Resolves #128


tom--pollard (approved):
LGTM, although I don't have much familiarity with golang repo structures. This won't be a breaking change to package-analysis as it's pinned but it's probably worth following it up.

Qinusty:
>This won't be a breaking change to package-analysis as it's pinned but it's probably worth following it up.

Will do this today :+1:

PR #293, opened 2023-02-21
third try to fix gcloud auth

maxfisher-g:


.github/workflows/auth-test.yml
```diff
@@ -0,0 +1,37 @@
+name: 'Test gcloud authentication'
+
+on:
+  pull-request:
```
oliverchang, line 4: Let's change this back to push now that we verified it's working.

.github/workflows/auth-test.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+
+    - name: Setup Environment
```
oliverchang, line 23: I think this step is no longer needed as well.

.github/workflows/auth-test.yml
maxfisher-g: Deleted this check action

.github/workflows/auth-test.yml
maxfisher-g: It's needed if we use the exported credentials file (see [here](https://github.com/google-github-actions/auth#other-inputs under `create_credentials_file`)

Can't tell if it's used though

oliverchang: `create_credentials_file` defaults to true, and it exports `GOOGLE_APPLICATION_CREDENTIALS` for us automatically right?

maxfisher-g: Yes, provided the checkout action is used above

maxfisher-g: Oh sorry, I forgot to push my latest change. I thought you were talking about the checkout action!

oliverchang (approved):

.github/workflows/auth-test.yml
```diff
@@ -0,0 +1,34 @@
+name: 'Test gcloud authentication'
```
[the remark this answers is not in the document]
maxfisher-g: It's deleted

PR #301, opened 2023-03-10
Add PyPI XML-RPC feed

### LINKED ISSUE #172: Consume PyPI updates from the XML-RPC mirroring api instead of RSS
PyPI updates can sometimes be missed.

To avoid this from happening the PyPI Mirroring API allows all changes to the PyPI package repository to be listed.

See: https://warehouse.pypa.io/api-reference/xml-rpc.html#mirroring-support

maxfisher-g:
Fixes #172 

This PR adds a new feed called `pypi-artifacts` that collects data from the [PyPI XML-RPC feed](https://warehouse.pypa.io/api-reference/xml-rpc.html#mirroring-support) rather than RSS. 

While this feed is officially deprecated, the current RSS feed has some problems with missing data when package archives are uploaded after the initial release, and the XML-RPC remains the only other fully supported way to reliably get this data.

Furthermore, a new optional field named `artifact_id` is added to the JSON schema; this allows identifying packages at the level of specific artifacts rather than just by the ecosystem/name/version triple.


pkg/feeds/pypi/pypi_artifacts.go
```diff
@@ -0,0 +1,143 @@
...
+		unixTimestamp = 0
+	}
+	action, ok := data[3].(string)
+	if !ok {
+		action = ""
+	}
+
+	archiveName := ""
+	// Changelog entries corresponding to new archives being added
+	// have an action string that looks like 'add <archive type> file <archive name>'
```
calebbrown, line 119: Might be worth referencing the code that generates this:
https://github.com/pypi/warehouse/blob/3bfd3e0d32e7396582d9635a316f1a47e407304d/warehouse/forklift/legacy.py#L1355 (added in https://github.com/pypi/warehouse/pull/863)

pkg/feeds/pypi/README.md
```diff
@@ -15,4 +16,20 @@ feeds:
     packages:
     - numpy
     - scipy
-```
\ No newline at end of file
+```
+
+# PyPI Artifacts Feed
```
calebbrown, line 21: This doc should include that the `artifact_id` is set to the filename.

calebbrown (changes requested):
Awesome! Looks great - just a couple of minor doc comments.

calebbrown (approved):

PR #302, opened 2023-03-13
update go to version 1.20

### LINKED ISSUE #300: Update go to latest version
The version of Go currently in use is 1.15, this should be updated to the latest version.

maxfisher-g:
Fixes #300


.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568
       with:
-        go-version: 1.15
+        go-version: 1.20
```
calebbrown, line 29: These need to be strings as YAML is turning them into number a number - which is why the tests are failing using v1.2 of Go rather than v1.20.

Use `go-version: '1.20'` instead

calebbrown (approved):

PR #312, opened 2023-04-06
Increase NPM feed limit, improve illegal XML character handling, remove utils.URLPathJoin 

### LINKED ISSUE #102: Improve npm polling performance

### LINKED ISSUE #171: Consume NPM updates from replicate.npmjs.com instead of RSS
https://replicate.npmjs.com/_changes is the CouchDB endpoint for consuming a feed of database changes. (see https://github.com/npm/registry/blob/master/docs/follower.md and https://docs.couchdb.org/en/3.2.0/api/database/changes.html)

The NPM RSS XML feed regularly has encoding issues:
```
time="2022-03-19T09:15:02Z" level=error msg="Error fetching packages" error="XML syntax error on line 69: illegal character code U+0000" feed=npm
```

Consuming the updates directly from the registry change API will:

1. ensure that encoding issues aren't encountered
2. ensure that *all* changes can be parsed

It isn't clear when new records are added, so there may be more updates in this feed than simply new packages. A filter may be required to limit results to only those we are interested in.

maxfisher-g:
- adds customisable limit to NPM RSS feed, which allows querying a greater number of packages at once
- rename `utils.UTF8OnlyReader` to `utils.XMLReader`
- fixes handling of illegal XML characters in RSS feeds by adding an option to `utils.NewXMLReader` (renamed from `utils.NewUTF8OnlyDecoder`) to replace non-XML characters with the □ character 
- replaces custom function `utils.URLPathJoin()` with library function `url.JoinPath` with the same semantics except in unused edge cases

Fixes #171

### DIFF
--- a/pkg/utils/utf8_decoder.go
+++ b/pkg/utils/utf8_decoder.go
@@ -8,7 +8,7 @@ import (
 )
 
 // UTF8OnlyReader implements a Reader which ignores non utf-8 characters.
-// Optionally, it can also replace non-XML characters with U+25A1 (□)
+// Optionally, it can also replace non-XML characters with U+25A1 (□).
 type UTF8OnlyReader struct {
 	buffer             *bufio.Reader
 	replaceNonXMLChars bool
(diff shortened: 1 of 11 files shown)


pkg/utils/utf8_decoder.go
```diff
@@ -7,33 +7,59 @@ import (
...
+		}
+
+		// Also ignore the replacement character for compatibility with previous behaviour
+		// (yes, utf8.RuneError == unicode.ReplacementChar)
 		if r == unicode.ReplacementChar {
 			continue
 		}
 
+		if reader.replaceNonXMLChars && !isInCharacterRange(r) {
+			r = '\u25A1' // □ (symbol for missing character)
```
calebbrown, line 57: Why can't this be `r = unicode.ReplacementChar`?

maxfisher-g: The choice of replacement character is arbitrary - perhaps I should make it an argument to the object. 

I chose not to use the unicode replacement character as I thought it was more often used for encoding errors, whereas this is a replacement of an unsupported character. But I can change it if you feel strongly

calebbrown: 25A1 is fine - perhaps specify it as a constant at the top, rather than here though, as it will make it clearer.

e.g. unicodeWhiteSquare = '\u25A1'

maxfisher-g: Fixed, please take a look

calebbrown (approved):

Issue #139, opened 2021-06-15
Sporadic error when polling for npm package data


tom--pollard:
`msg="Error fetching packages" error="Polling for package @deskpro/react-content-editor returned error: context deadline exceeded (Client.Timeout or context cancellation while reading body)" feed=npm`
https://registry.npmjs.org/@deskpro/react-content-editor

I've also logged it happening against https://registry.npmjs.org/@chainsafe/lodestar-utils & https://registry.npmjs.org/@duetds/css

As #123 introduced per package error handling the rest of the packages for the session as processed, in the past this would have resulted in a hard error & further data loss. The exact error isn't returned to the caller when triggered via the httpserver instead it's covered by the generic `pollErr` text.

Qinusty:
This could potentially be caused by rate limiting on requests as npm slows the rate of replies either intentionally or simply due to the increased load.

maxfisher-g:
Errors in log still occurring

calebbrown:
I think blindly generating one goroutine per npm package is a bit naive.

Instead, the packages should be pushed to a pool of workers that limit the amount of concurrency (e.g. instead of 200 reqs all trying to happen at once, only 10 can happen simultaneously.

calebbrown:
The errors are largely gone with the above changes.

The remaining timeouts appear to be limited to packages with ~1000 or more versions. I suspect increasing the timeout further (e.g. 60s) will likely prevent these issues.

maxfisher-g:
Closing as remaining timeout issues will be dealt with by #337

PR #330, opened 2023-05-03
Up the number of max idle conns allowed per host.

calebbrown:
By default Go sets the value to 2, which means keep-alive is only used for 2 connections, slowing down requests, and potentially causing unnecessary timeouts.


oliverchang (approved):

calebbrown:
This helps fix #139

PR #332, opened 2023-05-04
Add a pool of workers for fetching package from the NPM registry

calebbrown:
This limits the amount of simultaneous requests and should lower the
likelihood of errors being produced when querying NPM.

MaxIdleConnsPerHost is bumped to equal the number of workers.


calebbrown:
Ref #139

oliverchang (approved):

Issue #337, opened 2023-05-10
Large NPM package response data causes timeout


maxfisher-g:
After fixes for #139, remaining packages that still cause timeouts all have very large response data sizes (e.g. 6MB)

One solution may be to enable gzip compression during the request, as by default golang does not enable compression in the default http client. (It may also be possible to set up a roundtripper to do this automatically)

calebbrown:
Go appears to enable compression (gzip) by default on requests.

calebbrown:
Looking into this further, NPM supports HTTP2. I suspect that there is some weird behavior w.r.t timeouts and HTTP2 multiplexing.

calebbrown:
Inspecting the network traffic, there is only 1 connection being opened to `https://registry.npmjs.org/`, so yes, this is multiplexing queries over a single TCP+TLS connection to NPM.

This means that the large repos are congesting the single multiplexed HTTP2 connection. While the aggregate of the responses will be received faster than individual HTTP1 connections, each individual response is slower than if it was not multiplexed.

Some ideas for improving the performance:
- remove the limited workers, and just fire all the requests concurrently.
- add a small sleep (e.g. 0.1s) between each request to give the server some space.
- increase the timeout significantly (e.g. 1m30s) to allow for the slower responses.

maxfisher-g:
Thanks for investigating this @calebbrown!

calebbrown:
We still have errors here. Another attempt to reduce this will be to add an LRU cache and ETag "If-None-Match" checking on requets to NPM for a given package.

PR #339, opened 2023-05-11
Tune NPM timeouts and behavior for http2

calebbrown:
#337 

Tweak the behavior of NPM to work better with HTTP2 to reduce timeouts.

### DIFF
--- a/pkg/feeds/npm/npm.go
+++ b/pkg/feeds/npm/npm.go
@@ -204,8 +204,6 @@ func fetchAllPackages(registryURL string) ([]*feeds.Package, []error) {
 				packageChannel <- pkgs
 			}
 		}(pkgTitle, count)
-		// Pause briefly between each request so all the requests don't arrive at once.
-		time.Sleep(50 * time.Millisecond)
 	}
 
 	// Collect all the worker.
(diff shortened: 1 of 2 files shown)


pkg/feeds/npm/npm.go
```diff
@@ -176,70 +179,35 @@ func fetchAllPackages(registryURL string) ([]*feeds.Package, []error) {
...
+			// without being recorded in the json. An `event` could be logged
+			// here.
+			if len(pkgs) > count {
+				packageChannel <- pkgs[:count]
+			} else {
+				packageChannel <- pkgs
+			}
+		}(pkgTitle, count)
+		// Pause briefly between each request so all the requests don't arrive at once.
+		time.Sleep(50 * time.Millisecond)
```
maxfisher-g, line 208: nit: define the pause duration up in the constants

maxfisher-g (approved):
Just one suggestion, otherwise LGTM

pkg/feeds/npm/npm.go
calebbrown: Woops - I meant to delete that. I've added a random delay above inside the goroutine.

maxfisher-g (approved):

pkg/feeds/npm/npm.go
maxfisher-g: Haha that also solves it

PR #340, opened 2023-05-11
Add a threshold so small batches don't have jitter. Bump jitter window.

calebbrown:
- 10 seconds should space out requests more to reduce congestion.
- Early requests don't have jitter appled to avoid slowing down small batch sizes.

See #337

### DIFF
--- a/pkg/feeds/npm/npm.go
+++ b/pkg/feeds/npm/npm.go
@@ -185,10 +185,10 @@ func fetchAllPackages(registryURL string) ([]*feeds.Package, []error) {
 		uniquePackages[pkg.Title]++
 	}
 
-	n := 0
+	applyJitter := len(uniquePackages) > minJitterThreshold
 	for pkgTitle, count := range uniquePackages {
-		go func(id int, pkgTitle string, count int) {
-			if id > minJitterThreshold {
+		go func(pkgTitle string, count int) {
+			if applyJitter {
 				// Before requesting, wait, so all the requests don't arrive at once.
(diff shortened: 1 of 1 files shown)


pkg/feeds/npm/npm.go
```diff
@@ -179,11 +185,14 @@ func fetchAllPackages(registryURL string) ([]*feeds.Package, []error) {
 		uniquePackages[pkg.Title]++
 	}
 
+	n := 0
```
maxfisher-g, line 188: Can this be replaced by a boolean flag `applyJitter := len(uniquePackage) > minJitterThreshold`

pkg/feeds/npm/npm.go
```diff
@@ -179,11 +185,14 @@ func fetchAllPackages(registryURL string) ([]*feeds.Package, []error) {
...
 	for pkgTitle, count := range uniquePackages {
-		go func(pkgTitle string, count int) {
-			// Before requesting, wait, so all the requests don't arrive at once.
-			jitter := time.Duration(rand.Intn(maxJitterMillis)) * time.Millisecond //nolint:gosec
-			time.Sleep(jitter)
+		go func(id int, pkgTitle string, count int) {
+			if id > minJitterThreshold {
```
maxfisher-g, line 191: ...and this by `if applyJitter`? 

I think the intent would be clearer this way

maxfisher-g (changes requested):

pkg/feeds/npm/npm.go
calebbrown: Yes, good idea. Done

maxfisher-g (approved):
Great, thanks!

PR #341, opened 2023-05-11
Bump the memory limit as the job is running out of memory.

calebbrown:
Feeds is currently hitting the 512Mb limit regularly.


maxfisher-g (approved):

maxfisher-g:
Is this because of recent changes?

PR #347, opened 2023-05-21
Force HTTP1 for NPM and reintroduce a worker pool.

calebbrown:
HTTP2 appears to be struggling with the large responses from NPM and flow control.

Since this appears to be an issue inherent in HTTP2 and without wanting to delve into the complex task of tuning HTTP2 flow control parameters we fall back to pipelining in HTTP1.

### DIFF
--- a/pkg/feeds/npm/npm.go
+++ b/pkg/feeds/npm/npm.go
@@ -1,15 +1,16 @@
 package npm
 
 import (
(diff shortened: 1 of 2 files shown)


maxfisher-g (approved):
Would be good if we could factor out the parallel worker setup / logic from the package retrieval logic, perhaps in a future PR

PR #366, opened 2023-06-28
Improve support for cron (inc production)

calebbrown:
Upgrade cron library to v3.

Limit Cloud Run concurrency to "1" to avoid duplicate data being produced.


calebbrown:
On reviewing Cloud Run it appears to also be doing a health check on the same URL as the request to pull new packages.

Please hold off while I figure out a fix.

calebbrown:
The health check is on the port 8080 accepting TCP connections. I think it is okay (i.e. not causing work)

oliverchang (approved):

PR #370, opened 2023-06-28
Add an LRU cache for NPM responses to reduce HTTP traffic.

calebbrown:
This change introduces an LRU cache for NPM responses to avoid needless HTTP traffic for duplicate queries.

The hope is to improve the reliability of NPM requests by reusing cached data using the ETag in an "If-None-Match" header.

#337

### DIFF
--- a/pkg/feeds/npm/npm.go
+++ b/pkg/feeds/npm/npm.go
@@ -35,7 +35,7 @@ const (
 
 	// cacheEntryLimit defines how many responses to store in the LRU cache.
 	// The value should be larger than rssLimit to ensure all rss entries can
-	// be convered by a cache entry.
+	// be covered by a cache entry.
 	cacheEntryLimit = 500
 )
 
--- a/pkg/feeds/npm/npm.go
+++ b/pkg/feeds/npm/npm.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	lru "github.com/hashicorp/golang-lru/v2"
+
 	"github.com/ossf/package-feeds/pkg/events"
 	"github.com/ossf/package-feeds/pkg/feeds"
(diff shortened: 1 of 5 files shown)


pkg/feeds/npm/npm.go
```diff
@@ -52,9 +59,14 @@ type PackageEvent struct {
 	Title string `xml:"title"`
 }
 
+type cacheEntry struct {
+	ETag     string
```
maxfisher-g, line 63: could we add a little documentation? (yes this package is underdocumented)
e.g. what does the cache entry store exactly, where does the ETag come from, what does versions refer to?

pkg/feeds/npm/npm.go
```diff
@@ -153,15 +180,22 @@ func fetchPackage(httpClient *http.Client, baseURL, pkgTitle string) ([]*Package
 		return versionSlice[j].CreatedDate.Before(versionSlice[i].CreatedDate)
 	})
 
+	if etag != "" {
+		feed.cache.Add(versionURL, &cacheEntry{
```
maxfisher-g, line 184: nit: add comment explaining the implication of etag being blank - does it mean it's not in the cache and had to be retrieved?

maxfisher-g (dismissed):

maxfisher-g:
Looks good, just a little more documentation would be good for future code readers

pkg/feeds/npm/npm.go
calebbrown: Done

pkg/feeds/npm/npm.go
calebbrown: Done

maxfisher-g (approved):

maxfisher-g:
Thanks for the changes! Those are awesome comments :D

PR #383, opened 2023-08-16
Allow ScheduledFeed.Latest to return a cutoff for future invocations

### LINKED ISSUE #382: Replace lastPoll with a more robust way to track progress.
Currently when a feed is polled, the `FeedGroup.lastPoll` time is used as an argument to `feed.Latest` to specify the cutoff.

The value for `lastPoll` is updated after every feed in a feed group has been polled for latest packages.

This behavior has some serious drawbacks:
- if the FeedGroup has multiple feeds there will be a time gap between when the first feed is completed and when `lastPoll` is updated where updates may occur. This means that updates in the overlap may not be observed.
- the items in the feed may be delayed - e.g. all the items could be 5 minutes behind "now". This means that the cut off may cause updates to be ignored because the timestamp on the items is older than `lastPoll`.

A better approach is:
1. remove `FeedGroup.lastPoll` as an argument to `feed.Latest`.
2. have each feed define its own orderable approach to applying a cutoff.
3. the cutoff should be extracted from the feed itself (i.e. max value observed) and stored with the feed.

calebbrown:
This fixes #382 by ensuring each feed uses a cutoff value that is based on the data observed in the feed.

### DIFF
--- a/pkg/feeds/crates/crates_test.go
+++ b/pkg/feeds/crates/crates_test.go
@@ -31,6 +31,8 @@ func TestCratesLatest(t *testing.T) {
 	if len(errs) != 0 {
 		t.Fatalf("feed.Latest returned error: %v", err)
 	}
+
+	// Time should match the newest package creation time of packages retrieved.
 	wantCutoff := time.Date(2021, 3, 19, 13, 36, 33, 0, time.UTC)
 	if gotCutoff.Sub(wantCutoff).Abs() > time.Second {
 		t.Errorf("Latest() cutoff %v, want %v", gotCutoff, wantCutoff)
--- a/pkg/feeds/crates/crates_test.go
+++ b/pkg/feeds/crates/crates_test.go
@@ -27,9 +27,13 @@ func TestCratesLatest(t *testing.T) {
 	}
 
 	cutoff := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
-	pkgs, errs := feed.Latest(cutoff)
+	pkgs, gotCutoff, errs := feed.Latest(cutoff)
 	if len(errs) != 0 {
 		t.Fatalf("feed.Latest returned error: %v", err)
+	}
+	wantCutoff := time.Date(2021, 3, 19, 13, 36, 33, 0, time.UTC)
+	if gotCutoff.Sub(wantCutoff).Abs() > time.Second {
+		t.Errorf("Latest() cutoff %v, want %v", gotCutoff, wantCutoff)
 	}
 
 	if pkgs[0].Name != "FooPackage" {
@@ -67,7 +71,10 @@ func TestCratesNotFound(t *testing.T) {
 	}
 
 	cutoff := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
-	_, errs := feed.Latest(cutoff)
+	_, gotCutoff, errs := feed.Latest(cutoff)
+	if cutoff != gotCutoff {
+		t.Error("feed.Latest() cutoff should be unchanged if an error is returned")
+	}
 	if len(errs) == 0 {
 		t.Fatalf("feed.Latest() was successful when an error was expected")
 	}
--- a/pkg/feeds/crates/crates_test.go
+++ b/pkg/feeds/crates/crates_test.go
@@ -32,7 +32,7 @@ func TestCratesLatest(t *testing.T) {
 		t.Fatalf("feed.Latest returned error: %v", err)
 	}
 
-	// Time should match the newest package creation time of packages retrieved.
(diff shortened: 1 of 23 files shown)


pkg/feeds/pypi/pypi_test.go
```diff
@@ -27,10 +27,14 @@ func TestPypiLatest(t *testing.T) {
 	feed.baseURL = srv.URL
 
 	cutoff := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
-	pkgs, errs := feed.Latest(cutoff)
+	pkgs, gotCutoff, errs := feed.Latest(cutoff)
 	if len(errs) != 0 {
 		t.Fatalf("feed.Latest returned error: %v", err)
 	}
+	wantCutoff := time.Date(2021, 3, 19, 12, 1, 4, 0, time.UTC)
```
maxfisher-g, line 34: as above

pkg/feeds/packagist/packagist_test.go
```diff
@@ -29,13 +29,17 @@ func TestFetch(t *testing.T) {
...
 	cutoff := time.Unix(1614513658, 0)
-	latest, errs := feed.Latest(cutoff)
+	latest, gotCutoff, errs := feed.Latest(cutoff)
 	if len(errs) != 0 {
 		t.Fatalf("got error: %v", errs[len(errs)-1])
 	}
 	if len(latest) == 0 {
 		t.Fatalf("did not get any updates")
 	}
+	wantCutoff := time.Date(2021, 2, 28, 12, 20, 3, 0, time.UTC)
```
maxfisher-g, line 39: as above - is the unix time related here?

pkg/feeds/pypi/pypi_artifacts_test.go
```diff
@@ -403,10 +403,14 @@ func TestPyPIArtifactsLive(t *testing.T) {
 	}
 
 	cutoff := time.Now().AddDate(0, 0, -1)
-	pkgs, errs := feed.Latest(cutoff)
+	pkgs, gotCutoff, errs := feed.Latest(cutoff)
 	if len(errs) != 0 {
 		t.Fatalf("feed.Latest returned error: %v", err)
 	}
+	wantCutoff := time.Now().UTC()
+	if gotCutoff.Sub(wantCutoff).Abs() > 5*time.Minute {
```
maxfisher-g, line 411: why is it 5 minutes here when the other times are 1 second?

maxfisher-g (changes requested):
Code looks good, would be helpful to have comments describing the logic behind choices of `wantCutoff` times as well as length of error durations

pkg/feeds/feed.go
```diff
@@ -15,8 +15,15 @@ type UnsupportedOptionError struct {
 	Feed   string
 }
 
+// Cutoff is an interface used to define orering of package updates.
```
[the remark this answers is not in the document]
calebbrown: Removed the type as I was no longer using it.

pkg/feeds/feed.go
```diff
@@ -15,8 +15,15 @@ type UnsupportedOptionError struct {
...
+//
+// The type *time.Time satisifies this interface.
```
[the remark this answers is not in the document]
calebbrown: Removed the type.

pkg/feeds/pypi/pypi_artifacts_test.go
calebbrown: I fixed this. The test was pulling live data, so I now just check that it doesn't match the input `cutoff`.

pkg/feeds/packagist/packagist_test.go
calebbrown: Done.

The unix time above is not really related. I'm not sure why its used in this case.

pkg/feeds/crates/crates_test.go
```diff
@@ -27,10 +27,14 @@ func TestCratesLatest(t *testing.T) {
 	}
 
 	cutoff := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
-	pkgs, errs := feed.Latest(cutoff)
+	pkgs, gotCutoff, errs := feed.Latest(cutoff)
 	if len(errs) != 0 {
 		t.Fatalf("feed.Latest returned error: %v", err)
 	}
+	wantCutoff := time.Date(2021, 3, 19, 13, 36, 33, 0, time.UTC)
```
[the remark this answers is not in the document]
calebbrown: Done.

maxfisher-g (approved):

maxfisher-g:
Thanks for the changes! Great comments

PR #426, opened 2024-02-08
Expand parameters for container image building

JAORMX:
This adds the following:

* Expanded container metadata for the built image.
* multi-architecture builds
* Docker SBOM generation
* Provenance information generation

The intention is to keep a similar set of tags generated as before.

this also separates the pre-existing docker workflow into two jobs:

* One that builds the image on each PR
* Another one intended for the release.

This ensures that we get test coverage for image builds while keeping the GitHub actions permissions minimal.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,68 @@
+name: Release
+on:
+  push:
+    branches:
+      - main
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+env:
+  IMAGE_NAME: packagefeeds
+
+jobs:
+  # Push image to GitHub Packages.
+  push:
+    name: Push
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+
+    steps:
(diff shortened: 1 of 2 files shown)


lukehinds (approved):
A NB LGTM @JAORMX , but needs an approved commiter

@calebbrown are you able to review?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Build image
+        id: image-build
+        uses: docker/build-push-action@4a13e500e55cf31b7a5d59a38ab2040ab0f42f56 # v5
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: 
+          file: ./Dockerfile
+          tags: ${{ steps.docker-metadata.outputs.tags }}
+          labels: ${{ steps.docker-metadata.outputs.labels }}
```
calebbrown, line 69: How does the output of this differ when compared the previous implementation for the value of `$VERSION`?

JAORMX: should be quite similar as it uses the tag as well, and it strips the `v` prefix just using github action options. This adds even more metadata to the image.

calebbrown (approved):

PR #427, opened 2024-02-08
Add healthcheck

yrobla:
This is needed when deploying package-feeds on kubernetes, as it's best practice to have readiness and liveness endpoints. And we cannot just call the regular endpoint, because it actually triggers a publication of packages, it's better to have a dedicated endpoint for it.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -1,6 +1,7 @@
 package scheduler
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -36,6 +37,38 @@ type pollResult struct {
...
+		return
+	}
+
+	// Write an HTTP status code to the response
+	w.WriteHeader(http.StatusOK)
+
+	// Write the JSON response body
+	_, err = w.Write(jsonResponse)
+	if err != nil {
+		log.Errorf("Failed to write health check response: %v", err)
```
JAORMX, line 68: Should we just silently log the error? It seems to me like it would be kind of problematic if we're not able to write the healthcheck response. That could be indicative of other issues on the server.

calebbrown (approved):

PR #432, opened 2024-02-20
Add java support (maven repository)

yrobla:
Adds Maven support using Sonatype API, and specifically querying for Maven repository

### DIFF
--- a/pkg/feeds/feed.go
+++ b/pkg/feeds/feed.go
@@ -70,7 +70,7 @@ func NewArtifact(created time.Time, name, version, artifactID, feed string) *Pac
 func ApplyCutoff(pkgs []*Package, cutoff time.Time) []*Package {
 	filteredPackages := []*Package{}
 	for _, pkg := range pkgs {
-		if pkg.CreatedDate.UTC().After(cutoff) {
+		if pkg.CreatedDate.After(cutoff) {
 			filteredPackages = append(filteredPackages, pkg)
 		}
 	}
--- a/pkg/feeds/feed.go
+++ b/pkg/feeds/feed.go
@@ -70,7 +70,7 @@ func NewArtifact(created time.Time, name, version, artifactID, feed string) *Pac
 func ApplyCutoff(pkgs []*Package, cutoff time.Time) []*Package {
 	filteredPackages := []*Package{}
 	for _, pkg := range pkgs {
-		if pkg.CreatedDate.After(cutoff) {
+		if pkg.CreatedDate.UTC().After(cutoff) {
 			filteredPackages = append(filteredPackages, pkg)
 		}
 	}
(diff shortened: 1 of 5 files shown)


JAORMX (approved):

pkg/feeds/feed.go
```diff
@@ -70,7 +70,7 @@ func NewArtifact(created time.Time, name, version, artifactID, feed string) *Pac
 func ApplyCutoff(pkgs []*Package, cutoff time.Time) []*Package {
 	filteredPackages := []*Package{}
 	for _, pkg := range pkgs {
-		if pkg.CreatedDate.After(cutoff) {
+		if pkg.CreatedDate.UTC().After(cutoff) {
```
calebbrown, line 73: This looks like a good idea, how does this affect the current behavior?

Does it presume that each feed is using UTC?

pkg/feeds/maven/maven.go
```diff
@@ -0,0 +1,141 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/ossf/package-feeds/pkg/feeds"
+)
+
+const (
+	FeedName  = "maven"
```
calebbrown, line 14: This is a Maven Central feed, is it possible that we may want other Maven repository feeds in the future?

How would we distinguish those?

pkg/feeds/maven/maven.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error sending request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Handle rate limiting (HTTP status code 429).
+	if resp.StatusCode == http.StatusTooManyRequests {
+		time.Sleep(5 * time.Second)
+		return feed.fetchPackages(page) // Retry the request
+	}
```
calebbrown, line 79: This could repeat forever, it is probably worth adding some bound to how many retries are allowed.

calebbrown (commented):
Thanks for the contribution!

pkg/feeds/feed.go
yrobla: So I did that when i was comparing dates of maven with the cutoff. Cutoff was set to UTC, but i was seeing that the maven one was coming in CET (my local tz), so it was doing comparisons between CET and UTC. I assume that even with that, dates are compared properly as both include TZ information, but found more natural to compare with the same TZs.
I did not see any change of behaviour on the other feeds, tests passed and i can run it locally without problems. If you prefer i can remove this for now and we can do in another PR.

pkg/feeds/maven/maven.go
yrobla: Actually from Sonatype we can get more feeds apart from Maven. We are just filtering by repo name here. So it can be that we may have different feeds using same engine, but picking different repos. How do you see that, duplicating the feed and giving a new name?

calebbrown (approved):

calebbrown:
ref: #432

Issue #435, opened 2024-02-22
crates call being blocked by cloudflare


yrobla:
When ingesting the crates feeds i got 403 error consistently. When doing the same call from the laptop using curl, everything was working fine.
Debugged a bit more, and found that we were being blocked by Cloudflare.

PR #436, opened 2024-02-22
Fix crates

yrobla:
fix: add headers when querying crates
    
When running the package feeds for crates i was getting a 403 error, and when debugging response i was seeing that was being  blocked by Cloudflare.
    
Closes: #435

### DIFF
--- a/pkg/feeds/crates/crates.go
+++ b/pkg/feeds/crates/crates.go
@@ -41,7 +41,15 @@ func fetchPackages(baseURL string) ([]*Package, error) {
 	if err != nil {
 		return nil, err
 	}
-	resp, err := httpClient.Get(pkgURL)
+	req, err := http.NewRequest("GET", pkgURL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// set headers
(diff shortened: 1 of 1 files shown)


pkg/feeds/crates/crates.go
```diff
@@ -41,7 +41,15 @@ func fetchPackages(baseURL string) ([]*Package, error) {
...
 	}
-	resp, err := httpClient.Get(pkgURL)
+	req, err := http.NewRequest("GET", pkgURL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// set headers
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("User-Agent", "package-feeds")
```
calebbrown, line 51: Can we make the user agent `package-feeds (github.com/ossf/package-feeds)`?

Having the repository in the header gives administrators a point of contact if needed.

I have another change I'm preparing that adds this user-agent to all feeds, but I'm happy to proceed with this change to unblock you.

yrobla:
it's fine to just use your PR that is more general, i can abandon mine

Issue #386, opened 2023-08-28
Including an HTTP Publisher


R9295:
Hello all,

Any thoughts on including a HTTP publisher that sends a HTTP POST request to a configurable address? 
It is trivial to implement and if ya'll would like to introduce such functionality, I would love to contribute. 

Regards,
Aarnav

calebbrown:
Hi Aarnav,

You're definitely welcome to implement this feature. Web-hooks are a common approach (apart from pubsub) for receiving notifications when events occur.

Did you have in mind an approach for handling failures?


Caleb

R9295:
No I do not, but I imagine we can use the same strategies as the other publishers that work over the network?

yrobla:
I created https://github.com/ossf/package-feeds/pull/457

PR #457, opened 2024-03-26
feat: add new http publisher

yrobla:
Create a new publisher endpoint, capable of publishing content to an http feed.

Closes: #386


calebbrown (commented):
Thanks for this change!

After I give this a test run tomorrow or next week I'll merge it.

calebbrown (approved):

